PR #55, opened 2026-05-29
Drop JSON report from /conformu skill — text log is sufficient

joeytroy:
## Summary
- Removes the ConformU JSON report (\`conform.report.json\`) from the \`/conformu\` skill — both generation (\`-r\` flag in Step 3) and saving (the \`cp\` in Step 6b).
- The text log already carries everything the JSON gate was checking, so the JSON is purely redundant. Single source of truth, no drift risk.

## Why drop it
ConformU's text log emits two assertions that fully cover the JSON counts:

| JSON field | Text log assertion |
|---|---|
| \`ErrorCount\` / \`IssueCount\` / \`WarningCount\` = 0 | \`Congratulations, no errors, warnings or issues found\` |
| \`TimingIssuesCount\` = 0 | No \`OUTSIDE (FAST\|STANDARD\|EXTENDED) RESPONSE TIME TARGET\` lines (or the closing \`took longer than its target response time\` summary) |

The JSON's only edge over text was \"survives a future text-format change.\" After ConformU 4.3.0 the text format has been stable and predictable, and maintaining a parallel JSON copy creates more drift risk than it absorbs (already saw this with the JSON gate in /commit vs the text gate in /submit-pr both needing to be kept in sync).

## Changes (\`.claude/commands/conformu.md\` only)
- **Step 3** — drop \`-r \"\$TMPDIR/conform.report.json\"\` from the \`conformu\` invocation.
- **Step 4** — remove subsection 4a (\`jq\` on JSON counts). Renumber the text-log gates to 4a (pass message), 4b (timing greps), and the summary table to 4c. Update the summary to derive Errors / Issues / Warnings from the pass message line and Timing violations from grep.
- **Step 6b** — filenames list keeps only \`Linux-arm64.txt\` (and suffixed variants). Drop the \"BOTH the text log and the JSON\" guidance and the JSON \`cp\` line. Add a short paragraph explaining the text log is now the single source of truth.

## Compatibility
- \`/commit\` Step 3 and \`/submit-pr\` Step 1 ConformU hard-blocks already fall back to text-log checks when no JSON file is present, so they keep working on new runs with no edit needed.
- Existing committed JSON reports under \`AlpacaCore/conformu/<Vendor>/<Model>/Linux-arm64*.report.json\` are unaffected — this change is forward-only.

## Test plan
- [x] Skill markdown still well-formed (no broken anchors).
- [ ] Next \`/conformu\` run produces only \`Linux-arm64[-<transport>].txt\`, with the existing pass/fail gates working from the text log alone.

### DIFF
--- a/.claude/commands/conformu.md
+++ b/.claude/commands/conformu.md
@@ -133,8 +133,7 @@ Create a temp working directory and invoke the conformance subcommand:
 TMPDIR=$(mktemp -d /tmp/conformu-XXXXXX)
 <path-to-conformu>/conformu conformance \
   "http://<host>:<port>/api/v1/<type>/<n>" \
-  -n "$TMPDIR/conformu.txt" \
-  -r "$TMPDIR/conform.report.json"
+  -n "$TMPDIR/conformu.txt"
 ```
 
 Stream output to the user as it runs. ConformU can take minutes (especially for cameras and telescopes). Do not interrupt unless the user says so.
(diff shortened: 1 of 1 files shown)


joeytroy:
## Code Review — Drop JSON report from /conformu skill

**Must-fix issues: None.**

---

### Design notes

**1. ConfigurationAlertCount silently dropped**

The old summary table exposed `ConfigAlerts: N` as an informational (non-blocking) line. The new table has no equivalent — config alerts will be invisible unless the user reads the full log. Consider adding a line in the Step 4c summary template:

```
  Config alerts:               N            (scan for "ConfigurationAlert" in log; informational, not a gate)
```

This preserves visibility of a signal that may be useful to the driver author (e.g. "you didn't configure ASCOM profile defaults"), even though it doesn't affect the pass/fail verdict.

**2. FAIL-path summary is under-specified**

The `(from pass message)` annotation in the summary template implies the counts are derived by checking whether the pass message is present. That works cleanly for a pass. On a failure (pass message absent), there's no guidance on how to surface the actual error/issue/warning counts to the user. The old JSON approach gave exact numbers for free. Consider adding a note like:

> On FAIL: scan the text log for lines matching `Error …` / `Issue …` / `Warning …` per-member and summarise the tallies, since the pass-message-based count won't apply.

**3. Existing `.report.json` artifacts unaddressed**

There are likely already `Linux-arm64.report.json` (and transport-suffixed variants) committed to `AlpacaCore/conformu/`. The PR drops the instruction to save new ones but doesn't say what to do with the existing files. A brief backward-compat note would prevent confusion — e.g., "Existing `.report.json` files in the repo can be left in place; they are no longer generated for new runs."

---

### What's good

- The core simplification is sound: the `Congratulations, no errors, warnings or issues found` message is ConformU's own authoritative assertion — it is semantically equivalent to checking ErrorCount=0, IssueCount=0, WarningCount=0, so the JSON is genuinely redundant for pass/fail gating.
- The inline annotation explaining *why* the pass message is sufficient was a smart addition — future readers won't wonder whether the greps are a complete check.
- Removing the two-file drift problem (text log and JSON potentially diverging after a log format change) is a real improvement.
- The timing violation greps (§4b) are unchanged and remain authoritative.
- Step lettering (4a→4b→4c, 6b filenames) is updated consistently throughout.
- The defensive note "Do not generate or save a JSON report — the `-r` flag was dropped … precisely so the text log is the single source of truth" is good defensive documentation.

Overall a clean, well-motivated change. Items 1–3 above are worth a follow-up but none are blockers for merging.

---

PR #56, opened 2026-05-29
Add ZWO ASIair Pro 12V power Switch driver (ConformU 4.3.0)

joeytroy:
## Summary
- Adds an ASCOM Switch driver for the ZWO ASIair Pro that exposes the four on-board 12V DC power ports via on-board Linux GPIO. Each port is independently configurable as a boolean on/off channel or a 0–100% software PWM channel (for dew heaters and flat panels).
- Built on **libgpiod v2** (the modern character-device API), not pigpio, so the driver is portable across arm64 SBCs and doesn't fight a daemon for line ownership.
- ConformU 4.3.0 validated on Linux arm64: **0 errors, 0 issues, 0 timing issues** with a mixed config (2 boolean + 2 PWM ports). Slowest member 21 ms vs the STANDARD 1 s target.

## Changes
- **ZWO ASIair Pro Switch Driver** (AlpacaCore): new \`SwitchDriver\` implementation; per-port boolean-or-PWM mode; default Pi 4 ASIair Pro layout (Port 1=GPIO 12, Port 2=GPIO 13, Port 3=GPIO 26, Port 4=GPIO 18 on \`/dev/gpiochip0\`); per-port worker threads handle soft-PWM (configurable Hz, default 1 kHz).
- **ASIair GPIO Protocol Wrapper** (AlpacaCore): 4th-file \`zwo_asiair_protocol_wrapper\` that owns all four lines via a single \`gpiod_line_request\` and serializes line writes; clean teardown on disconnect.
- **ZWO ASIair Device Support** (AlpacaHTTP): router accepts \`vendor=zwo deviceType=switch switchType=asiair\` with optional \`gpioChip\`, \`pwmFrequencyHz\`, and \`ports[]\` overrides for non-Pi-4 deployments. Web UI Switch Type dropdown gains the new option and shows a per-port table (channel name, GPIO line, PWM checkbox) plus GPIO chip path and PWM frequency inputs. Edit-flow restores all saved fields from the persisted device config.
- **ZWO ASIair Unit Tests**: 9 test cases / 74 assertions covering defaults, device metadata, per-port metadata (channel ranges, descriptions, default names), \`NotConnected\`/\`InvalidValue\` error codes, switch ID range validation, state machine when disconnected, unsupported method error codes, and constructor rejection of invalid configs.
- **ConformU Validation**: \`AlpacaCore/conformu/ZWO/ASIair Pro/Linux-arm64.txt\` + \`.report.json\` saved. AB TRACE log scanned for \`WARN/ERROR/exception/retry/timeout/reconnect\` across 2,793 lines → zero matches.
- **Packaging** (debian/control): \`libgpiod-dev (>= 2.0)\` added to Build-Depends, \`libgpiod3\` to runtime Depends.
- **Documentation**:
  - \`AlpacaCore/PowerPorts.md\` — new copy-paste end-user setup guide (apt deps, \`gpio\` group membership, boot config note, Web UI walk-through, \`gpioget\` verification). Stubbed sections for ASIair Plus (Pi CM4 / RK3568), ToupTek StellaVita, iOptron iMate to be filled in as those validate.
  - \`SUPPORTED-DRIVERS.md\` — ASIair Pro 12V Power row updated from "Pending re-image" + "ConformU pending" to ✓ + ConformU link; Driver Notes appended. Updated date 2026-05-18 → 2026-05-29.
  - \`AGENTS.md\` — comprehensive ZWO ASIair Pro section: hardware reality (Pi 4 + BCM2711), the actual port-to-GPIO mapping with a note that the \`/boot/config.txt\` order is **not** the port order, the runtime persistent-state schema discovered by SSH inspection of the stock app, the role-vs-channel cosmetic distinction, why we use libgpiod v2 over pigpio's DMA PWM, and the OS architecture gate (stock ASIair is armv7l so deployment requires re-imaging to Pi OS 64-bit).
  - \`CHANGELOG.md\` — new bullets under 1.0.4 UNRELEASED.

## Test plan
- [x] Unit tests pass: 51 cases / 360 assertions full suite, 9 cases / 74 assertions for \`[asiair]\` filter, on arm64
- [x] ConformU 4.3.0 passes on Linux arm64 with mixed boolean+PWM config: 0 errors, 0 issues, 0 timing issues
- [x] Web UI configuration works in browser (Configure + Edit flows both populate the per-port table correctly)
- [x] Device connects and switches toggle in NINA (verified Port 1–4 boolean on/off and PWM 0–100 round-trip)
- [x] Hardware: physical ASIair Pro re-imaged to Debian 13 Trixie arm64, kernel 6.18.29+rpt-rpi-v8

## ConformU results
- **arm64**: [\`AlpacaCore/conformu/ZWO/ASIair Pro/\`](AlpacaCore/conformu/ZWO/ASIair%20Pro/)
  - \`Linux-arm64.txt\` — 369-line ConformU 4.3.0 log
  - \`Linux-arm64.report.json\` — \`{"ErrorCount":0,"IssueCount":0,"TimingIssuesCount":0}\`


joeytroy:
## Code Review — ZWO ASIair Pro 12V Power Switch Driver

**ConformU 4.3.0: 0 errors, 0 issues, 0 timing violations** on Linux arm64, mixed 2-boolean + 2-PWM config. Clean pass.

---

### Must-fix issues: None

No correctness bugs that block functionality or safety.

---

### Design notes

**1. PWM worker thread silently ignores `gpiod_line_request_set_value` failures**

The PWM loop writes to the GPIO line without checking the return value:

```cpp
::gpiod_line_request_set_value(request, offset, GPIOD_LINE_VALUE_ACTIVE);
std::this_thread::sleep_for(std::chrono::microseconds(on_us));
```

If the GPIO write fails (hardware fault, kernel error), the thread loops silently — the ASCOM API keeps accepting `SetSwitchValue` calls and reporting success while the hardware isn't responding. The thread should log the error on failure and exit via `stop->store(true)`. Since the thread can't throw, logging + break is the right action. The non-PWM boolean path correctly catches errors via the mutex-protected `gpiod_line_request_set_value` check.

**2. GPIO line may remain HIGH after disconnect (hardware pull-ups)**

When `close()` is called while a PWM worker is in the HIGH phase of its duty cycle:

```cpp
::gpiod_line_request_set_value(request, offset, GPIOD_LINE_VALUE_ACTIVE);
std::this_thread::sleep_for(std::chrono::microseconds(on_us));
if (stop->load()) {
    break;  // exits WITHOUT setting the line LOW first
}
```

The thread exits leaving the line HIGH. When the `gpiod_line_request` is released, the kernel reverts the line to its default state. Because the ASIair Pro hardware has pull-ups (`gpio=18,12,13,26=op,dh,pu` in `/boot/firmware/config.txt`), a released line likely stays HIGH. Result: **gear stays powered after the driver disconnects** — potentially surprising for a power switch driver.

The fix is to drive all lines LOW before releasing the request in `close()`:

```cpp
// in close(), after workers exit and before gpiod_line_request_release:
for (std::size_t i = 0; i < ports_.size(); ++i) {
    ::gpiod_line_request_set_value(request_, ports_[i].gpio_line, GPIOD_LINE_VALUE_INACTIVE);
}
```

Alternatively, document explicitly that disconnecting leaves ports in their last state (and may leave them powered). AGENTS.md captures the boot-time default-on behavior but not the disconnect behavior. Either fix the behavior or document it — the current gap between expectation ("disconnect the driver = safe") and behavior ("ports stay on") is a user safety concern for a 12V power switch.

**3. Negative GPIO number accepted by router, silently wraps to large uint32_t**

In `router.cpp`:
```cpp
if (!p.contains("gpio") || !p["gpio"].is_number_integer()) {
    error_message = "ASIair port entry requires integer 'gpio'";
    return false;
}
pc.gpio_line = static_cast<std::uint32_t>(p["gpio"].get<int>());
```

`is_number_integer()` passes for `-1` (or any negative integer). The cast wraps to `4294967295`, which libgpiod rejects at connect time with `EINVAL` — not a crash, but the error message will be confusing. Add `|| p["gpio"].get<int>() < 0` to the guard (and optionally `|| p["gpio"].get<int>() > 63` as a sanity upper bound).

**4. CMake libgpiod fallback doesn't verify ≥ 2.0**

```cmake
else()
    find_library(LIBGPIOD_LIBRARY NAMES gpiod libgpiod)
    find_path(LIBGPIOD_INCLUDE_DIR NAMES gpiod.h)
    if(LIBGPIOD_LIBRARY AND LIBGPIOD_INCLUDE_DIR)
        # links without version check
```

The pkg-config path correctly enforces `>= 2.0`, but the fallback `find_library` path would happily link libgpiod 1.x. The v2 API functions (`gpiod_chip_request_lines`, `gpiod_line_request_*`) don't exist in v1, so the build would fail with cryptic linker/compiler errors. Consider adding a version check in the fallback path or using `find_package` with a VERSION argument. Not a blocker on Trixie (which ships libgpiod 2.x), but will confuse developers on older Bullseye/Bookworm hosts.

**5. PWM thread wakes every `period_us` at 0% and 100% duty**

At `d <= 0` or `d >= 100`, the thread continuously drives the line and sleeps `period_us`. At 1 kHz this is 1000 kernel ioctls per second for a static output. A small optimization (check if the target hasn't changed since the last write and skip the `gpiod_line_request_set_value` call) would reduce kernel overhead for steady-state high/low ports. Minor, not a blocker.

**6. `Linux-arm64.report.json` and PR #55**

PR #55 (currently open, not yet merged) proposes dropping `.report.json` going forward. This PR was clearly authored before #55 was finalized, and the JSON file is explicitly included. Per PR #55's compatibility note, existing JSON files are unaffected — this is fine either way. If #55 merges before #56, this PR's JSON is just the last one ever generated. Not actionable for this PR.

---

### What's good

- **pimpl pattern** cleanly hides all libgpiod types from the header — no gpiod.h in the public interface.
- **All five special member functions** correctly deleted (copy ctor/assign + move ctor/assign + dtor) — prevents accidental resource sharing.
- **Exception path in `open()`** manually frees `req_cfg`, `line_cfg`, `settings`, and the chip in the catch block before rethrowing. All four gpiod resources are freed on every failure path.
- **`close()` mutex dance** is correctly written: lock → set stop flags → **unlock before join** (avoids deadlock) → join workers → relock → release GPIO → set `open_ = false`. Clean pattern.
- **Boolean `set_value` acquires the mutex** before calling `gpiod_line_request_set_value`, correctly serializing with the PWM threads (which use the same request but different line offsets — serialized at the kernel level, but the mutex prevents a boolean write racing with a `close()`).
- **`validate_id` checks `< 0` before cast** — correctly catches negative switch IDs before the unsigned cast.
- **`set_switch` delegates to `set_switch_value`** — keeps the on/off mapping (0/1 for boolean, 0/100 for PWM) in a single place.
- **Unit tests** cover the full disconnected API surface cleanly without requiring hardware — `NotConnected`, `InvalidValue`, `ActionNotImplemented`, `NotImplemented`, range rejection, per-port metadata, name persistence, and config validation (empty ports, zero freq, out-of-range freq).
- **Web UI `dispatchEvent(new Event('change'))`** in `startEditDevice` is the correct way to trigger the section visibility update when loading an existing asiair-type device — this handles the edit flow correctly.
- **ConformU tested with a mixed config** (2 boolean + 2 PWM ports) so both code paths were exercised in a single run. Slowest member 21 ms vs 1 s STANDARD target — wide margin throughout.
- **SUPPORTED-DRIVERS.md** and **AGENTS.md** both updated comprehensively with implementation-side context that will help future contributors.
- **`debian/control`** correctly splits `libgpiod-dev (>= 2.0)` into Build-Depends and `libgpiod3` into runtime Depends.

---

joeytroy:
## Review response — all six notes addressed in db84863

Mapping each item from the review to the resolution:

### #1 — PWM worker silently ignored gpiod_line_request_set_value failures → **fixed**
The PWM worker now error-checks every `gpiod_line_request_set_value` call. On failure it logs at ERROR with the GPIO offset + errno + a clear "aborting worker thread" message, then sets the per-port `stop` flag so the thread exits cleanly. No more silently looping while the ASCOM API reports success.
`AlpacaCore/src/vendors/zwo/zwo_asiair_protocol_wrapper.cpp`

### #2 — Disconnect leaves lines HIGH → **documented** (intentional behavior)
After thinking it through, this is the right behavior: it mirrors the boot-time `gpio=18,12,13,26=op,dh,pu` default-on policy, and avoids surprising users with an unexpected power cycle on every transient ASCOM disconnect. The fix is documentation, not code change:

- `AlpacaCore/PowerPorts.md` — new "Disconnect behavior — important for unattended observatories" section explaining the policy and the toggle-off-before-disconnect pattern for users who want a port OFF after disconnect.
- `AGENTS.md` — explicit warning for future contributors: do **not** add a "drive LOW on close" path without making it opt-in via config, because it would silently flip the policy for everyone who upgrades.

### #3 — Negative GPIO accepted by router → **fixed**
Router now explicitly rejects `gpio < 0` and `gpio > 63` with `"ASIair port 'gpio' must be in [0, 63]"` before the cast. No more silent wrap to a large uint32_t followed by a cryptic libgpiod `EINVAL` at connect.
`AlpacaHTTP/src/http/router.cpp`

### #4 — CMake fallback didn't verify libgpiod ≥ 2.0 → **fixed**
The `find_library` fallback path now probes for `gpiod_chip_request_lines` (v2-only symbol) via `CheckSymbolExists` before accepting the library. If missing, CMake fails with a clear message naming the symbol, the required version, and the apt package to install. No more cryptic linker errors on Bullseye/Bookworm hosts that have libgpiod 1.x.
`AlpacaCore/src/vendors/zwo/CMakeLists.txt`

### #5 — PWM thread does 1000 kernel ioctls/sec for static output → **fixed**
The PWM worker now caches the last-written value. Steady-state 0% / 100% duty ports skip the ioctl entirely when the desired value matches the cached value. Mid-duty paths invalidate the cache (the line genuinely toggles every period). First write still happens because `last_written` starts at `-1`. Same file as #1 — the error-handling and caching share the same `write_line` helper lambda for clean code.
`AlpacaCore/src/vendors/zwo/zwo_asiair_protocol_wrapper.cpp`

### #6 — JSON report vs PR #55 → **no action needed**
You noted PR #55 makes future `/conformu` runs text-only and that this PR's JSON is grandfathered. Confirmed; nothing to do.

---

### What still passes
- All 51 AlpacaCore tests / 360 assertions
- All 9 ASIair-specific tests / 74 assertions (the `[asiair]` filter)
- ConformU artifact on disk is unchanged — the runtime API surface didn't move, the fixes harden internal paths and input validation. A re-run would produce the same 0/0/0/0 result.

Ready for another pass.

joeytroy:
Second pass on `db84863` — all five notes addressed correctly.

**#1 PWM error handling**: The `write_line` lambda checks the `gpiod_line_request_set_value` return on every call, logs at ERROR with the GPIO offset + errno string + "aborting worker thread", then sets `stop->store(true)` and returns false. Both the 0%/100% steady-state paths (via `write_line`) and the mid-duty ACTIVE/INACTIVE writes are now covered. Clean.

**#2 Disconnect behavior documented**: AGENTS.md captures the policy clearly — "do **not** add a 'drive LOW on close' path without making it opt-in via config, because it would silently flip the policy for everyone who upgrades." The reasoning is sound: mirroring the boot-time default-on policy is the right default for an observatory setting; an unexpected power cycle mid-session is worse than gear staying on. The PowerPorts.md "Disconnect behavior" section gives users the pattern for opt-out (set the port OFF in the client before disconnecting). Good call on documentation over code change here.

**#3 GPIO range**: `gpio_value` is now read into a signed `int` local first, range-checked against `[0, 63]` with a clear error message, and only then cast to `uint32_t`. No more silent wrap. ✓

**#4 CMake fallback**: `CheckSymbolExists(gpiod_chip_request_lines "gpiod.h" LIBGPIOD_V2_OK)` is exactly the right probe — it's a v2-only function that v1 doesn't declare, and the `FATAL_ERROR` message names the symbol, the version requirement, and the apt command. ✓

**#5 PWM caching**: `last_written = -1` forces the first write; the steady-state paths skip ioctl when `desired == last_written`; mid-duty correctly invalidates the cache with `last_written = -1` before toggling. The single `write_line` helper handles both fixes (#1 and #5) cleanly in one lambda. ✓

**#6**: Acknowledged, nothing to do.

LGTM — no remaining issues.

---

PR #57, opened 2026-05-29
Speed up build_and_run.sh: incremental by default, ccache + Ninja auto

joeytroy:
## Summary
- Rewrites \`build_and_run.sh\` so it is **incremental by default**: a no-op rerun finishes in ~5 sec (\`ninja: no work to do.\`) instead of 5–10 min, and a one-file edit rebuilds in 10–30 sec instead of 5–10 min. Same \`./build_and_run.sh\` invocation, no new flags.
- Auto-installs \`ccache\` + \`ninja-build\` via apt the first time they are missing (uses the same sudo escalation as the existing udev-rule install block) and auto-uses both when present. Configurable via \`ALPACA_INSTALL_ACCELERATORS=ON|OFF\` (default ON).
- Removes a dead \`ALPACAHTTP_USE_BOOST_BEAST\` CMake option that was warning on every configure.

## Changes
- **\`build_and_run.sh\`**:
  - **Stops wiping \`AlpacaHTTP/build\`** on every run. CMake + Ninja are incremental by design; only files that actually changed get recompiled. \`CLEAN=1 ./build_and_run.sh\` is the escape hatch for forcing a full rebuild.
  - **Drops the inner \`make clean\`** call inside the build loop. That single line was the actual killer for incremental builds even when the build dir survived.
  - **Drops the standalone AlpacaCore build** entirely. AlpacaHTTP pulls AlpacaCore via \`add_subdirectory(../AlpacaCore)\`, so the standalone \`libalpacacore.a\` was never used at runtime.
  - **Auto-installs ccache + ninja-build** via \`sudo apt-get install -y\` when missing. The check is a microsecond \`command -v\` per run; the install only happens once per fresh system.
  - **Auto-uses Ninja** as the CMake generator and **ccache** as the C/C++ compiler launcher whenever both binaries are present.
  - **Auto-cleans on generator switch.** If the user previously ran the old Make-based script and Ninja is now installed, the script detects the mismatch via \`CMakeCache.txt\` and wipes the dir once with a one-line message instead of failing with a cryptic CMake error.
  - **Caches udev-rule + firmware + vendor \`.so\` install** behind a hash-stamp file at \`AlpacaHTTP/build/.system_setup_stamp\`. The stamp is keyed off every input the install block would consume; any vendor SDK update (new \`.so\`, new rules file, new firmware blob) invalidates it automatically, but a no-change rerun skips the entire sudo-prompting block.

- **Dead-option cleanup** (\`build_and_run.sh\` + \`AlpacaHTTP/README.md\`):
  - Drops \`-DALPACAHTTP_USE_BOOST_BEAST=...\` from the cmake invocation. The option is not declared anywhere in \`AlpacaHTTP/CMakeLists.txt\` (no \`option()\`, no \`add_definitions\`, no source guard), so CMake emitted \`Manually-specified variables were not used by the project: ALPACAHTTP_USE_BOOST_BEAST\` on every configure.
  - Drops the matching bullet from \`AlpacaHTTP/README.md\`'s Build Options list — it was misleadingly listed with a \"default: ON\" description for a feature that doesn't exist.

- **\`CHANGELOG.md\`**: bullet under 1.0.4 UNRELEASED → Changed documenting the speedup and the new env-var escape hatches.

## Measured impact (arm64, M4 dev box; Pi 4 numbers similar)

| Scenario | Before | After |
|----|----|----|
| Fresh clean build (Ninja + ccache cold) | 5–10 min | **~4 min** |
| No-op re-run, no source changed | 5–10 min | **~5 sec wall, 0.13 s user CPU** (\`ninja: no work to do.\`) |
| Touch a single \`.cpp\` file + rerun | 5–10 min | **~10–30 sec** depending on dependents |
| Branch switch back to previously-compiled state | 5–10 min | **near-instant** (ccache hits) |

## How it handles \`git pull\` (i.e. when new code lands)

The script doesn't track changes — **Ninja and CMake do**. Ninja maintains a per-file mtime/hash database; on each run it figures out which \`.cpp\` files changed, which \`.h\` files changed (and recompiles every \`.cpp\` that transitively included them), and which \`CMakeLists.txt\` files changed (triggering a re-configure). ccache adds a third layer — even when a file does need recompiling, an unchanged preprocessed result is served from the cache. New vendor SDK files automatically invalidate the install stamp.

The only edge cases that need \`CLEAN=1\` are toolchain upgrades (e.g. g++ 14.2 → 14.3) and confused linker errors after big cross-branch refactors. Both rare.

## Test plan
- [x] No-op rerun verified on arm64: \`ninja: no work to do.\`, total runtime dominated by the server start.
- [x] Full clean build (Ninja + ccache cold) verified: 4m0s wall on M4 dev box.
- [x] CMake reconfigure after the dead-option removal emits zero warnings.
- [x] Auto-install of ccache + ninja-build verified on a host that was missing both.
- [x] Generator-switch auto-clean verified (Make → Ninja transition).
- [ ] Run on a fresh Raspberry Pi 4 with stock Pi OS 64-bit to confirm full first-run flow (apt install + udev + build) works end-to-end.


joeytroy:
## Code Review — Incremental build_and_run.sh

**CI: N/A** (shell script, no automated CI for this path)

---

### Must-fix issues

None.

---

### Design notes

**1. Stamp hash uses size+path, not content**

```bash
find "${CORE_DIR}/external" -type f ... -printf '%s %p\n' 2>/dev/null | sort | sha256sum
```

`%s` is file size. A vendor SDK update that happens to produce a `.so` of exactly the same byte count won't invalidate the stamp. For vendor SDK blobs this is an acceptable trade-off (version bumps almost always change size), but worth knowing — `CLEAN=1` is the workaround if someone hits it.

**2. `ALPACACORE_BUILD_TESTS=OFF` addition**

Good catch — the old standalone AlpacaCore build was never consumed at runtime, and now that AlpacaCore is pulled only via `add_subdirectory`, passing `-DALPACACORE_BUILD_TESTS=OFF` to the AlpacaHTTP configure command prevents AlpacaCore's test suite from being compiled as part of every server build. Correct.

**3. Generator auto-clean — `cut -d= -f2` assumption**

```bash
existing_generator="$(grep '^CMAKE_GENERATOR:' ... | head -1 | cut -d= -f2)"
```

`CMakeCache.txt` stores this as `CMAKE_GENERATOR:INTERNAL=Ninja` — `cut -d= -f2` gives `Ninja`. This works as long as the generator name contains no `=`. All current CMake generator names (`Ninja`, `Unix Makefiles`, `Ninja Multi-Config`) are `=`-free, so this is fine in practice.

**4. `set -e` + stamp write ordering**

If `run_system_setup` fails mid-way, `set -e` kills the script before `echo "${setup_hash}" > "${SETUP_STAMP}"` runs. The stamp is never written, so the next run retries the full setup. This is the correct failure behavior — a half-completed install won't be stamped as clean.

---

### What's good

- Dropping `make clean` inside the build loop is the right fix — that was the actual 5–10 min killer per run, not the CMake configure step.
- Generator-mismatch auto-clean is a nice UX touch; the "was Make, now Ninja" transition would otherwise produce an opaque CMake error.
- The `PARALLEL` simplification (three-way if/elif/else replacing nested ifs) is cleaner.
- `CLEAN=1` removes the whole `build/` dir including the stamp, so system setup re-runs — documented in the header comment. Correct and consistent.
- Dead `ALPACAHTTP_USE_BOOST_BEAST` option correctly excised from both the script and `README.md`.

LGTM ✅

---

PR #58, opened 2026-05-30
Add ToupTek filter wheel HTTP integration (router + web UI)

zeroastroboy:
Adds the missing HTTP layer and web UI support for the ToupTek filter wheel driver.

### DIFF
--- a/AlpacaHTTP/src/http/router.cpp
+++ b/AlpacaHTTP/src/http/router.cpp
@@ -79,6 +79,7 @@
 #endif
 #ifdef ALPACACORE_ENABLE_TOUPTEK
 #include <alpacacore/vendor/touptek/touptek_camera_driver.h>
+#include <alpacacore/vendor/touptek/touptek_filterwheel_driver.h>
 #include <alpacacore/vendor/touptek/touptek_focuser_driver.h>
 #endif
 #ifdef ALPACACORE_ENABLE_PLAYERONE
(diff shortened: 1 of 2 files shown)


joeytroy:
## Code Review — ToupTek Filter Wheel HTTP Integration

Thanks for this contribution, @zeroastroboy! The implementation is clean and follows the existing router registration pattern well. A few notes:

### What looks good

**Router pattern is correct.** The `vendor == "touptek" && device_type_str == "filterwheel"` guard, the `#ifdef ALPACACORE_ENABLE_TOUPTEK` compile-time gate, and the `registry.register_device(...)` return-value check all mirror the existing ToupTek camera path exactly. The error messages for both the "not compiled in" and "already registered" cases are consistent with the rest of the codebase.

**`cameraIndex` for filter wheel selection is the right abstraction.** ToupTek integrates the filter wheel into the camera body, so using the camera index to identify which wheel to control is the logical mapping. The `config.value("cameraIndex", 0)` default-to-0 is a safe fallback.

**Web UI changes are scoped correctly.** Enabling ToupTek in the filter wheel vendor dropdown, preventing a reset of the ToupTek selection when switching to filter wheel, and showing the `cameraIndex` field for filter wheels are all consistent with how the camera side works. The JS changes are minimal and targeted.

### Questions / things to verify

1. **`create_touptek_filterwheel` existence.** The router calls `alpacacore::vendor::touptek::create_touptek_filterwheel(device_number, camera_index)` — can you confirm this factory function exists in AlpacaCore and has this exact signature? There's no CI running for this project, so a compilation check isn't automated; if you've built it locally against the current AlpacaCore, that confirmation would be helpful in the PR description.

2. **`cameraIndex` config key casing.** The camera registration path uses `"cameraIndex"` (camelCase) and so does this PR — that's consistent. Just worth noting so future contributors don't introduce a `camera_index` variant.

3. **No new tests** — understood for a hardware-driver integration, but if there's a mock/stub path in the test suite it might be worth a note.

Otherwise this looks ready to merge pending confirmation of point 1. Nice addition!

---

PR #59, opened 2026-05-31
Add ZWO ASIair Plus (RK3568) Switch driver — ConformU 4.3.0 ✓

joeytroy:
## Summary

Adds the ZWO ASIair Plus (Rockchip RK3568 variant) Switch driver. Exposes the four on-board 12 V DC ports as ASCOM Switch channels with per-port boolean on/off or userspace soft-PWM at a configurable frequency. Validated end-to-end with ConformU 4.3.0: **0 errors, 0 issues, 0 timing violations**.

This is a sister driver to the existing ASIair Pro (Pi 4) Switch — the kernel-interface differences (closed ZWO `pwm_gpio.ko` ioctl vs. libgpiod v2) were fundamental enough to warrant a separate driver and protocol wrapper.

## Changes

- **ZWO ASIair Plus (RK3568) Switch Driver** (AlpacaCore): new `SwitchDriver` + 4th-file `AsiairPlusProtocolWrapper` that opens `/dev/pwm-gpio-misc` and issues the reverse-engineered `pwm_gpio.ko` ioctl interface (header vendored at `AlpacaCore/external/ZWO/asiair-plus/pwm_gpio.h`). Wrapper indices 0–3 map to kernel ioctl indices 4–7 (DC1–DC4). PWM-marked ports each run a userspace soft-PWM worker thread toggling `SET_LEVEL` at the configured frequency (default 200 Hz, range 1–100 000 Hz). The kernel module's hrtimer hardware-PWM dispatch branch turned out unreachable from any documented ioctl sequence, and GPIO bank 4 has no PWM peripheral mux on the RK3568 — independently confirmed against the extracted stock `zwoair_imager` daemon binary, which takes the same software-PWM approach. `SET_LEVEL` polarity is inverted from typical gpiod conventions (`level=0` ⇒ pad floats high ⇒ panel ON; `level=1` ⇒ output low ⇒ panel OFF); the wrapper translates this so ASCOM `value=1` always means on. Connect-time policy is read-only: `open()` opens the fd and writes nothing — the first hardware write is the user's first `SetSwitch` / `SetSwitchValue` call.
- **ZWO ASIair Plus Device Support** (AlpacaHTTP): router accepts `switchType: "asiair-plus-rk3568"` with `devicePath`, `pwmFrequencyHz`, and `ports[]` overrides. Sanitization allowlist updated so per-port config (`ports[]`, `pwmFrequencyHz`, `devicePath`) actually persists through save → reload cycles for both ASIair variants. Web UI gains a new Switch Type option "ASIair Plus 12V Power Switch (RK3568)" with a simpler per-port table (channel name + PWM checkbox; no GPIO field since the kernel module fixes the mapping).
- **ZWO ASIair Plus Unit Tests**: 9 test cases / 75 assertions covering defaults, device metadata, per-port metadata, `NotConnected`/`InvalidValue` error codes, switch ID range validation, state machine when disconnected, unsupported method error codes, and constructor rejection of invalid configs.
- **ZWO ASIair Plus udev rule** (`AlpacaCore/external/ZWO/asiair-plus/99-zwo-asiair-plus.rules`): grants the `gpio` group `0660` access to `/dev/pwm-gpio-misc`. Installed automatically by `build_and_run.sh`, which now also auto-creates the `gpio` group on stock Debian (it doesn't exist by default) and adds the daemon user to it.
- **PowerPorts.md ASIair Plus (RK3568) section**: promoted from stub to full setup guide.
- **ConformU 4.3.0 validation**: 0 errors / 0 issues / 0 timing violations on Linux arm64 (Debian 13 Trixie, kernel 4.19.219 + stock ZWO `pwm_gpio.ko`). Mixed config (3 boolean + 1 PWM) exercised both code paths in a single run. Slowest member `SetSwitch (3)` at 30 ms — 3 % of the STANDARD 1 s target.
- **Documentation**: CHANGELOG 1.0.4 UNRELEASED (Added/Changed/Removed entries); SUPPORTED-DRIVERS.md Switch table row → ✓ with validation link, Driver Notes corrected to match shipping behavior; AGENTS.md ZWO vendor section captures the full reverse-engineering forensics (read-only open() policy, GET_LEVEL-returns-cache-not-pad-state caveat, SET_MODE+ENABLE+SET_LEVEL sequence requirement, polarity inversion at SET_LEVEL, kernel-module symbol-table notes).
- **Orphaned ConformU JSON report removed**: the Pro driver's `Linux-arm64.report.json` deleted to match the recent `/conformu` policy change — text log is the single source of truth.

## Test plan

- [x] Unit tests pass — `cd AlpacaCore/build && ./tests/alpacacore_tests "[asiair-plus]"` → 75 assertions in 9 test cases pass
- [x] ConformU 4.3.0 passes on Linux arm64 — 0 errors / 0 issues / 0 timing violations, log committed
- [x] Web UI configures the new switch type end-to-end (verified in browser: add device, mark PWM, save, reconnect, set values via NINA)
- [x] Device connects and operates correctly on real hardware (ASIair Plus running Debian 13 Trixie + stock ZWO kernel 4.19.219)

## ConformU results

- **arm64**: [AlpacaCore/conformu/ZWO/ASIair Plus (RK3568)/Linux-arm64.txt](AlpacaCore/conformu/ZWO/ASIair%20Plus%20(RK3568)/Linux-arm64.txt)


joeytroy:
## Code Review — PR #59 (ZWO ASIair Plus RK3568 Switch Driver)

The implementation itself looks solid — the ioctl sequence, polarity inversion, `sleep_until` PWM timing, two-phase close, and ConformU 4.3.0 clean pass are all correct. There are **four stale documentation bugs** where comments contradict the actual code. These will mislead contributors who read the docs without reading the source.

---

### Bug 1 — `zwo_asiair_plus_protocol_wrapper.h`: class-level comment lies about `open()`

```cpp
// Master enable (kernel index 3) is driven HIGH on open() so the per-port
// switches downstream can deliver voltage. On close() we leave each port in
// its last-set state — mirroring the Pi-4 ASIair Pro disconnect policy.
class AsiairPlusProtocolWrapper { ... };
```

The `.cpp` `open()` implementation is fully read-only — it only opens the fd and reads `GET_LEVEL` / `GET_CONFIG` for each port, with an explicit long comment explaining *why* no write ioctls are issued at open time (power-cycling DC ports on every daemon start is obviously wrong). The master-enable is driven HIGH only by the first `apply_port_locked` call when a port is actually turned on. The comment needs to match the implementation.

**SUPPORTED-DRIVERS.md** (added in this same PR) gets it right — it doesn't claim any init-time master-enable toggle. Update the header comment to match.

---

### Bug 2 — `AGENTS.md` line 24: wrong claim about worker threads

```markdown
the RK3568 driver does **not** spawn worker threads. The kernel module's
hrtimer handles the period+duty toggling directly
```

Both of these statements are false. `open()` calls `std::thread([this, port_idx, ...]{ pwm_loop(...); })` for every PWM-enabled port — those threads are joined in `close()`. The kernel hrtimer path is unreachable from any documented ioctl sequence (SUPPORTED-DRIVERS.md, added in this same PR, correctly calls it "userspace soft-PWM worker thread"). Fix this before another contributor tries to debug a thread issue by looking at the kernel module.

---

### Bug 3 — `CHANGELOG.md` Added entry: two wrong claims

```markdown
Added: ZWO ASIair Plus (RK3568) switch driver — hardware PWM via the kernel hrtimer …
       master-enable (kernel index 3) auto-driven HIGH on connect
```

- **"hardware PWM via the kernel hrtimer"** — it's userspace soft-PWM (`pwm_loop` threads, `sleep_until`). The kernel hrtimer branch is unreachable (your own SUPPORTED-DRIVERS.md explains why).
- **"master-enable (kernel index 3) auto-driven HIGH on connect"** — `open()` is read-only; see Bug 1.

Changelogs live forever in the history and tend to be quoted verbatim in downstream docs. Worth correcting now.

---

### Bug 4 — `app.js`: comment claims 20 kHz default, code uses 200 Hz

```js
// pwmFrequencyHz was previously surfaced here as a user-editable field.
// It's now auto-managed by the wrapper (defaults to 20 kHz for inaudible soft-PWM)
```

`default_asiair_plus_rk3568_config()` sets `pwm_frequency_hz = 200`. SUPPORTED-DRIVERS.md (this PR) correctly documents "default **200 Hz** — empirically tuned as the quietest stable point on real hardware (1 kHz audible whine band, ≤ 100 Hz scheduler-jitter flicker)". The comment is 100× off and directly contradicts the reasoning in the same PR. Delete or correct it.

---

### Everything else looks good

- Ioctl sequence (`SET_MODE(GPIO)` → `ENABLE` → `SET_LEVEL`) matches the reverse-engineered header.
- Polarity inversion (`value != 0 ? 0 : 1`) is correct per your `GET_KEYS` analysis.
- Two-phase close (signal under mutex, join outside mutex, close under mutex again) correctly avoids deadlock.
- `sleep_until` in `pwm_loop` prevents cumulative drift.
- `std::lround` before range-check in `set_switch_value` is correct.
- `sanitize_device_config` allowlist in `router.cpp` properly covers the plus-variant fields.
- `ensure_gpio_group` in `build_and_run.sh` is idempotent and safe.
- 9 test cases / 75 assertions provide solid coverage; no gaps noticed.
- ConformU 4.3.0 — 0 errors, 0 issues, 0 timing violations.


---

joeytroy:
## Follow-up — all four doc fixes confirmed ✅

Bugs 1–4 from my earlier review are all fixed in the new commit. The header class comment, AGENTS.md, CHANGELOG.md, and app.js are now consistent with the implementation.

---

### New clarity issue — PowerPorts.md PWM limitation note is ambiguous

The note at the top of the ASIair Plus section reads:

> **Limitation: PWM dimming is not yet functional on stock Debian.** … PWM-enabled ports therefore behave as **always-on** until this is resolved.

But based on AGENTS.md and the implementation, the driver uses **userspace soft-PWM via GPIO mode (mode=1 + `pwm_loop` threads toggling `SET_LEVEL`)**, not the kernel's mode=2 PWM path. The note appears to describe an investigation of the kernel's mode=2 path (which genuinely doesn't produce output), but a reader will take away "the AlpacaBridge PWM feature is broken."

The comment block around the 200 Hz default (lines ~1688–1696 in the cpp) describes audible frequency testing at 1 kHz, 500 Hz, 200 Hz — which implies the soft-PWM was tested on real hardware and *does* produce duty-cycle control. If that's the case, the "always-on" note is only true for the kernel's native mode=2, not for the driver's own soft-PWM path.

Two concrete changes that would prevent user confusion:

1. **Clarify what "not functional" refers to.** Something like: "The kernel module's built-in PWM mode (`SET_MODE(PWM)`) is unreachable and produces no output. The driver avoids it entirely — PWM dimming is implemented via userspace soft-PWM (`pwm_loop` threads, GPIO mode) and does produce actual duty-cycle control."

2. **Line 177 says "empirically validated against the ZWO kernel hrtimer from 100 Hz to 100 kHz (every readback exactly matched…)"** — that's a CONFIG-register echo test, not an output test. A reader seeing this right after the "not functional" note will be confused: either the hrtimer works (line 177) or it doesn't (line 86). Suggest replacing with: "The driver's soft-PWM accepts any value in 1–100,000 Hz; the kernel's mode=2 CONFIG register echoes back any frequency you write (GET_CONFIG matches SET_CONFIG), but does not produce actual output."

Not a blocker, but the note as written will generate support questions from the first user who configures a PWM port and sees "always-on" — they'll think dimming is broken when in fact it should be working.


---

joeytroy:
## Follow-up 2 — PowerPorts.md clarity issues resolved ✅

Both issues from my second comment are addressed in the latest commit:

**The "not functional" ambiguity is gone.** The new blockquote clearly separates the two paths:
> *We attempted [SET_MODE(PWM)] first and found it a dead end … the AlpacaBridge driver bypasses SET_MODE(PWM) entirely and instead spawns a per-port worker thread … Verified end-to-end against real hardware — both boolean toggle AND duty-cycle dimming produce the expected physical behavior. Tick the PWM checkbox freely; PWM works.*

**The CONFIG-register echo is correctly described.** Line 177's "validated against the ZWO kernel hrtimer" claim is now a parenthetical aside in the Advanced configuration section explicitly calling it out as CONFIG-register storage, with "the kernel's hrtimer never arms and the pad never toggles."

All review items resolved. Implementation is correct, all four original doc bugs are fixed, and the documentation now accurately represents what the driver does and doesn't do. **LGTM — ready to merge.**


---

PR #72, opened 2026-06-05
Add iOptron iMate PowerBox Switch driver (PWM, ConformU validated)

joeytroy:
## Summary
- New ASCOM **Switch** driver for the **iOptron iMate PowerBox** DC power ports (iMate = OrangePi 3 LTS / Allwinner H6, arm64), driven over local GPIO via **libgpiod v2** — independent of the iOptron mount RS-232 protocol.
- Exposes 3 switches: **DC3** always-on read-only pass-through, **DC1/DC2** controllable. DC1/DC2 can each opt into **0–100% soft-PWM dimming** for dew heaters and flat panels.
- Runs on the OpenAstro Armbian (mainline-kernel) image; main GPIO bank is `/dev/gpiochip1`.
- **ConformU 4.3.0 on real iMate hardware: 0 errors, 0 issues, 0 timing issues.**

## Changes
- **iOptron iMate PowerBox Driver** (AlpacaCore): dedicated `ioptron_powerbox_wrapper` (libgpiod v2) + `IoptronSwitchDriver` (ISwitchV3). DC1 = line 118 (PD22), DC2 = line 114 (PD18), DC3 = no GPIO / read-only. Per-port soft-PWM via a worker-thread bit-bang (default **50 Hz**, 1–100 kHz), PWM port presented as an analog 0–100 channel. Never cuts power on disconnect (boolean holds its level; a PWM line is driven to a defined steady level before release).
- **iOptron Device Support** (AlpacaHTTP): router accepts `vendor=ioptron deviceType=switch` with `gpioChip`, `pwmFrequencyHz`, and a positional `ports[]` overlay (per-port `pwm`/`name`); config sanitization persists them. Web UI adds iOptron as a Switch vendor with GPIO-chip, PWM-frequency, and DC1/DC2 PWM-checkbox fields.
- **Unit + Routing Tests**: 18 iOptron switch cases / 124 assertions; AlpacaHTTP routing test covers registration + `gpioChip`/`pwmFrequencyHz`/`ports` persistence + `MaxSwitch`.
- **ConformU Validation**: Linux arm64 (OpenAstro mainline kernel, `/dev/gpiochip1`), mixed config (DC1 PWM, DC2 boolean, DC3 read-only) so all three port types ran in one pass — 0/0/0.
- **Documentation**: CHANGELOG, SUPPORTED-DRIVERS.md (Switch row → ✓), AGENTS.md (iMate PowerBox notes incl. the PWM frequency findings + OpenAstro/`gpiochip1` migration), and a full PowerPorts.md iMate setup section.

## Test plan
- [x] CI pre-flight green: build+test (vendors OFF + ON), clang-format, unicode, clang-tidy, cppcheck, JS check, sanitizers (ASan+UBSan)
- [x] Unit tests pass (124 assertions / 18 cases)
- [x] ConformU 4.3.0 passes on Linux arm64
- [x] Web UI configuration works (device add + PWM toggles)
- [x] Device connects and operates: DC1/DC2 toggled and PWM-dimmed on real hardware

## ConformU results
- **arm64**: `AlpacaCore/conformu/iOptron/iMate PowerBox/Linux-arm64.txt` (+ `.report.json`) — 0 errors, 0 issues, 0 timing issues.


joeytroy:
Thanks for the review — all three addressed in 21ff253, plus the clang-tidy CI failure:

**1. Breaking build for telescope-only / non-GPIO hosts** — fixed. The iMate PowerBox Switch sources + libgpiod requirement are now gated on libgpiod (>= 2.0) being found; without it the iOptron **mount** driver still builds and the switch is disabled with a CMake `STATUS` message (no more `FATAL_ERROR`). Propagated via a cache variable → `ALPACACORE_IOPTRON_POWERBOX` macro so the router branch and the switch unit test gate to match. Verified both configs build + test (libgpiod present, and forced-absent to exercise the telescope-only path).

**2. Inconsistent default PWM frequency** — fixed. `IoptronPowerboxWrapper`'s constructor now defaults `pwm_frequency_hz` to **50 Hz**, matching `IoptronSwitchConfig`.

**3. Asymmetric disconnected-write in `set_value()`** — fixed. The PWM path now applies the same `!open_` → `NotConnected` guard as the boolean path, so the wrapper is self-contained.

**clang-tidy CI failure** — the job installed apt `libgpiod-dev` (v1.6) and configured vendors-OFF, so it parsed the v2 vendor sources against v1 `<gpiod.h>` (spurious `clang-diagnostic-error`s). It now builds libgpiod v2 from source into `/usr/local`, mirroring the `build-vendors` job. Local pre-flight (incl. zizmor, since `ci.yml` changed) is green.

joeytroy:
Addressed in a643908:

- **Supply chain (libgpiod checksum)** — both the clang-tidy and `build-vendors` download steps now verify the tarball against kernel.org's published sha256 (`13207176…9592`) via `sha256sum -c` before extracting.
- **`stop.store(false)` in `close()`** — added a comment explaining the re-arm (supports close-then-reopen).

On the two informational minors, I'm leaving them as-is since the effective enforcement point already covers them, but happy to tighten if you'd prefer:
- *`ports` extra keys / `pwmFrequencyHz` range at sanitization*: an out-of-range frequency (or a malformed port) is rejected when the device is registered — the `IoptronPowerboxWrapper` constructor throws `InvalidValue`, so configuration fails and nothing operates on a bad value. The sanitizer is intentionally a pass-through allowlist; the constructor is the validation gate.

Local pre-flight green incl. zizmor (`No findings`) on the workflow change.

joeytroy:
Re: the "truncated SHA-256 (63 chars)" blocking note — this is a miscount; the digest is correct and full-length. Evidence:

- It's **64 hex characters**: `printf '%s' 13207176b0eb9b3e0f02552d5f49f5a6a449343ce47416158bb484d9d3019592 | wc -c` → `64`.
- It matches kernel.org's published `sha256sums.asc` for `libgpiod-2.2.4.tar.xz` exactly (`13207176…3019592`).
- **CI proves it executes correctly**: both the `build-vendors` and `clang-tidy` jobs run `sha256sum -c` against the downloaded tarball, and both are green on this commit. A 63-char digest would make `sha256sum -c` fail with *"no properly formatted checksum lines found"* and those jobs would go red.

So no change needed there.

On the two informational notes:
- *Connection failure swallowed on `open()`* — matches the existing ZWO ASIAIR pattern (logged, `Connected` stays false); a project-wide behavior rather than something new here.
- *`close()` latency at very low PWM frequency* — fair; at the 1 Hz floor a worker can take up to ~half a period to observe `stop`. Happy to add a one-line note if you'd like, but it doesn't affect the 50 Hz default (~10 ms).

PR #77, opened 2026-06-10
Add ASCOM Platform 7 interface versions and compliant DeviceState

joeytroy:
## Summary
- Advertise Platform 7 interface versions across all drivers: ICameraV4, ITelescopeV4, IFocuserV4, IRotatorV4, IObservingConditionsV2 (FilterWheel/Switch already reported IFilterWheelV3/ISwitchV3)
- Replace per-vendor DeviceState overrides with one spec-compliant base-class implementation per device type (standard property names + ISO 8601 TimeStamp)
- ZWO ASI290MM Mini validated with ConformU 4.3.0 on Linux arm64 at ICameraV4: 0 errors, 0 issues, 0 timing issues

## Changes
- **Platform 7 DeviceState** (AlpacaCore): base-class `get_device_state()` builds the property list by calling the device's own getters in try/catch, so DeviceState always agrees with the matching GET endpoints (the consistency ConformU verifies); a shared inline `device_state_timestamp()` appends the TimeStamp; inline definitions keep per-vendor static libraries free of link-order dependencies. Intentional empty catches are annotated `NOLINT(bugprone-empty-catch)` per-site rather than disabling the check project-wide.
- **Unit Tests**: per-driver InterfaceVersion assertions updated; DeviceState regression test added (TimeStamp present, no `Connected`/`CoolerOn`); full AlpacaCore suite green (179 tests, vendors ON)
- **ConformU Validation**: ASI290MM Mini through a full exposure/ImageArray cycle plus the async Connect/Disconnect lifecycle
- **Documentation**: CHANGELOG.md, SUPPORTED-DRIVERS.md (ASI290MM Mini row → validated), AGENTS.md lessons learned

## Test plan
- [x] Local CI pre-flight green: clang-format, unicode scan, build+tests (vendors OFF and ON), clang-tidy, cppcheck, ASan+UBSan sanitizers
- [x] ConformU 4.3.0 passes on Linux arm64 (ASI290MM Mini)
- [x] ConformU V4 hardware re-validation of telescope/focuser/rotator/ObservingConditions/FilterWheel drivers still pending before the 2.0.0 release

## ConformU results
- **arm64**: `AlpacaCore/conformu/ZWO/ASI/ASI290MM Mini/Linux-arm64.txt`


joeytroy:
Re: review feedback — addressed in 29d259f.

**Narrow exception catch** — agreed and fixed. All six base-class `add()` lambdas now catch `std::exception` (which `AlpacaException` derives from via `std::runtime_error`), so an unwrapped vendor SDK exception degrades to an omitted property instead of propagating out of `get_device_state()`. Added a regression test (`ThrowingFocuser` in `test_device_registry.cpp`) covering both a raw `std::runtime_error` and an `AlpacaException`.

**`TrackingRate`** — not applied: it is not part of the ITelescopeV4 DeviceState property set. The authoritative member list in the ASCOM Library ([`TelescopeState.cs`](https://github.com/ASCOMInitiative/ASCOMLibrary/blob/main/ASCOM.Common/ASCOM.Common.DeviceStateClasses/TelescopeState.cs)) is exactly: Altitude, AtHome, AtPark, Azimuth, Declination, IsPulseGuiding, RightAscension, SideOfPier, SiderealTime, Slewing, Tracking, UTCDate, TimeStamp. The removed iOptron implementation's `TrackingRate` entry was one of the non-standard members this PR deliberately eliminates; re-adding it would reintroduce a non-spec property name.

**Snapshot atomicity** — acknowledged as a behavior change; the ASCOM spec doesn't require an atomic snapshot, and per-getter reads are what guarantee the DeviceState-vs-GET-endpoint consistency that ConformU actually checks. The Celestron alt/az TTL cache keeps the serial traffic bounded.

joeytroy:
Second-pass review addressed in c478bde (docs only):

- **AGENTS.md catch-type mismatch** — fixed; it now documents `catch (const std::exception&)` matching the code.
- **Non-atomic snapshots** — documented in AGENTS.md (with guidance not to add a snapshot lock, and not to build on mutual consistency within one DeviceState response) and noted in the CHANGELOG entry. As the review notes, the ASCOM spec doesn't require atomicity and ConformU only checks DeviceState ↔ GET consistency.

joeytroy:
Third-pass review verified against the code — no changes needed this round:

- **iOptron cache refresh** — the individual getters refresh their own caches under the driver mutex: `get_altitude()`/`get_azimuth()` call `refresh_altaz_cache_locked()`, `get_at_home()`/`get_at_park()` call `refresh_status_cache_locked()` (ioptron_telescope_driver.cpp:394-441). The base-class DeviceState therefore never reports stale values; the old explicit refresh calls in the removed override were redundant with what the getters already do.
- **iOptron `AtPark` override guard** — the `park_override_until_` time guard lives in `get_at_park()` itself (ioptron_telescope_driver.cpp:426-435), not in the removed DeviceState path, so the override window is preserved through the new getter-delegation path.
- **Switch DeviceState** — not a gap: switch DeviceState is intentionally per-vendor (the property set is dynamic — `GetSwitchN`/`GetSwitchValueN`/`StateChangeCompleteN` per switch ID). All five switch drivers (iOptron iMate, ToupTek StellaVita, ZWO ASIair Pro/Plus, ZWO) override `get_device_state()`, and it's hardware-validated — e.g. the iMate PowerBox ConformU log: "Received 9 operational state properties … Found all expected operational properties" (0 errors/issues).
- **`%lld` on older MSVC** — moot: AlpacaBridge targets Linux arm64 only (no Windows builds; amd64 is dropped in the upcoming 2.0.0).

PR #79, opened 2026-06-11
Unify web UI header with guider; generate debian/changelog

joeytroy:
## Summary
- Finishes the web UI's visual unison with OpenAstro Guider: natural brand word spacing, header bar at the guider's ~57px height, version badge restyled as a guider pill (`v1.0.3`, dim monospace), and the tagline removed
- Ports the guider's deb-changelog design (guider commit ce7c5598): `debian/changelog` becomes a generated build artifact derived from `CHANGELOG.md`, so release history is maintained in exactly one place — the tracked copy was frozen at a 1.0.0 stub while `VERSION` says 1.0.3
- Fixes an unsatisfiable build dependency found while verifying: `debian/control` demanded `libcurl4-openssl-dev` specifically, but the code uses flavor-agnostic `find_package(CURL)`; now accepts the gnutls flavor too

## Changes
- **Web UI header** (AlpacaHTTP): `.title-text` wrapper keeps "OpenAstro AlpacaBridge" in inline flow (the flex `gap` was inflating the word space); header padding tuned so the bar renders at the guider's ~57px; version badge in guider pill styling with `v` prefix; tagline and its unused `.subtitle` rule removed
- **Debian packaging**: new `scripts/changelog_to_deb.py` (guider's generator adapted to this repo's `## [X.Y.Z] - UNRELEASED` heading and `VERSION` file, with mismatch warnings) and `scripts/build_deb.sh` (the repo's first package build entry point: generate → `dpkg-parsechangelog --all` validation → `dpkg-buildpackage -us -uc -b`); `debian/changelog` untracked and gitignored along with `*.deb`/`*.buildinfo`/`*.changes`
- **debian/control**: `libcurl4-openssl-dev | libcurl4-gnutls-dev`
- **Documentation**: CHANGELOG web-UI entry amended to the final header state plus a new packaging entry; AGENTS.md Debian Packaging section documents the generated-changelog flow

## Test plan
- [x] Local CI pre-flight green: unicode scan, build+tests (vendors OFF and ON), shellcheck, `node --check`
- [x] Header verified live in headless Chromium: 57px bar (guider: 57.3px), badge `v1.0.3`, computed pill color identical to the guider's (`rgb(139, 150, 168)`)
- [x] Packaging verified end-to-end: `scripts/build_deb.sh` produced `alpacabridge_1.0.3_arm64.deb` with all 28 release stanzas in `usr/share/doc/alpacabridge/changelog.gz`; every stanza parses under `dpkg-parsechangelog`


joeytroy:
All three review points addressed in the latest commit:

- **UTC stanza dates** — `section_date()` now builds the datetime with `tzinfo=datetime.timezone.utc` directly; verified output shows `12:00:00 +0000`. (This makes our port stricter than the guider original, which has the local-noon behavior.)
- **Version format guard** — early `VERSION_RE` check; `--version feature/foo` now exits with `error: --version 'feature/foo' is not a valid Debian version`.
- **Hardcoded arm64 in done message** — now `$(dpkg --print-architecture)`. Worth noting it was unreachable-wrong: `debian/rules` hard-fails on any non-arm64 host, so the message could never print on x86-64 — but the dynamic form is correct regardless.

The two minor observations (continuation-line fold, empty-bullet guard) are acknowledged as non-blocking; this repo's CHANGELOG format doesn't hit either, and the generator is validated by `dpkg-parsechangelog --all` on every build.

PR #99, opened 2026-07-01
Add ToupTek AFW-M filter wheel + ATR2600M cooled-camera support

joeytroy:
## Summary
- **ToupTek AFW-M filter wheel** (standalone, 5/7-slot) via the toupcam SDK — homes the wheel at connect (read slot count → write back → reset), matching the INDI toupbase reference, which is what stops it hunting after a firmware flash.
- **ToupTek ATR2600M cooled-camera support** — conversion gain (HCG/LCG/HDR) + High Full Well as ASCOM `ReadoutModes`, `Offset` (black level), a **Thermal Switch** (dew heater / fan / tail LED), and a fix for a 3×3-binning hang.
- All three devices **ConformU 4.3.0 validated on Linux arm64: 0 errors, 0 issues, 0 timing issues.**
- Cuts the **2.2.0** release (VERSION, README badge, CHANGELOG dated).

## Changes
- **ToupTek FilterWheel Driver** (AlpacaCore): `touptek_filterwheel_driver`, IFilterWheelV3; homing at connect; single absolute moves; `Names`/`FocusOffsets` shared semantics.
- **ToupTek Camera** (AlpacaCore): conversion gain + High Full Well folded into `ReadoutModes`; `Offset` via `TOUPCAM_OPTION_BLACKLEVEL`; **3×3 binning fix** — odd bin factors produced an odd sensor-ROI span that `Toupcam_put_Roi` rejected (ImageReady never set); ROI span now rounded to even.
- **ToupTek Thermal Switch** (AlpacaCore): new switch driver (dew heater / fan / tail LED), capability-probed; shares the camera's Toupcam handle via a reference-counted open so it runs alongside the camera device.
- **ToupTek SDK Wrapper** (AlpacaCore): filter-wheel + CG / high-full-well / black-level / heat / fan / tail-light options; reference-counted `open_camera_by_id`/`close_camera`.
- **ToupTek Support** (AlpacaHTTP): router registration for the filter wheel and thermal switch (`switchType: thermal` vs `stellavita`); web UI slot picker, Switch Type selector, and index auto-numbering.
- **Unit Tests**: filter wheel (8 cases, 43 assertions), thermal switch (8 cases, 43 assertions), plus camera readout-mode & offset cases.
- **Documentation**: CHANGELOG (2.2.0), SUPPORTED-DRIVERS.md, docs/architecture.md, AGENTS.md.

## Test plan
- [x] Local CI pre-flight green (build+tests vendors OFF/ON, clang-format, clang-tidy, cppcheck, unicode, JS, ASan+UBSan sanitizers)
- [x] Unit tests pass
- [x] ConformU 4.3.0 passes on Linux arm64 (all three devices)
- [x] Web UI configuration works
- [x] Devices connect and operate correctly on hardware (ATR2600M + AFW-M)

## ConformU results (Linux arm64, 0/0/0 each)
- **AFW-M filter wheel**: `AlpacaCore/conformu/ToupTek/AFW-M/`
- **ATR2600M camera**: `AlpacaCore/conformu/ToupTek/ATR2600M/`
- **ATR2600M thermal switch**: `AlpacaCore/conformu/ToupTek/ATR2600M Thermal Switch/`

## Note for reviewers
The first camera ConformU run threw a transient "Connected returned True after disconnect" that **did not reproduce** on the clean re-run (the saved report). It's the shared async-disconnect poll handler (`router.cpp`), not this driver — flagging it for awareness.


joeytroy:
Round-19 review addressed in e5c2c2f:

- **#1 `exposure_active_` off `readout_mutex_`** — the exposure thread's own exit clears (abort/failed + natural completion) now publish the false-transition under `readout_mutex_`, completing the stated invariant. Taken alone there (thread holds neither `mutex_` nor the SDK lock; no join site holds `readout_mutex_`), so lock order is preserved and it can't deadlock the join.
- **#3 `close_shared` underflow** — now only decrements when the ref count is `> 1` and treats `<= 1` as the final release, so a mispaired double-close can't underflow past the erase and `Toupcam_Close` a still-referenced handle.
- **#2 disconnect dropped during the connect window** — as noted, this is the shared async-connect pattern across ~26 drivers, so a one-off fix here would create the site-asymmetry our review checklist warns against. Tracked for the shared `AsyncConnectable` base in #100 (cancellation-aware connect loop) rather than patched into this driver.

Full sanitizer pre-flight (ASan+UBSan) green.

joeytroy:
Round-20 review addressed in ac4dec5. Two of the suggested fixes I implemented differently, on purpose:

- **#1 ImageReady/Exposing ordering** — fixed as suggested: the exposure thread now clears `exposure_active_` (leaving Exposing) before publishing `image_ready_`, with the frame built/cached while still Exposing so the flip window carries no heavy work.
- **#2 `reset_exposure_state_locked` off `readout_mutex_`** — fixed as suggested, at both the redundant-reconnect and post-connect reset sites, so every `exposure_active_` transition is now under `readout_mutex_`.
- **#3 `set_switch` TOCTOU** — fixed via the *single-lock resolve+write* option, **not** the "write 1.0/0.0 directly" suggestion: the dew heater and fan are ranged (max = `HEAT_MAX` / `maxfanspeed`, not 1), so a hardcoded 1.0 would only ever set them one step above off. A shared `write_element_value_locked` helper now does resolve+range+write under one lock.
- **#4 `close_shared` double-close** — the suggested `if (handle && id_by_handle_.find(handle) == end()) return;` would have **leaked every `open_camera_by_index` handle** (those are intentionally absent from `id_by_handle_` and must still close). Instead a live-handle set makes `close_shared` idempotent: a no-op for handles not currently open, while untracked index handles still close.

Full sanitizer pre-flight (ASan+UBSan) green.

joeytroy:
Round-21 addressed in 7c44e08:

- **enumerate_cameras null-model guard** — real fix: added `if (!arr[i].model) continue;` so a null-model entry can't slip past the `arr[i].model && …` short-circuit and register as a phantom camera at index 0. Now matches `enumerate_focusers`/`enumerate_filter_wheels`.
- **open_camera_by_index index space** — documented that it's the SDK's raw enumeration order (includes AFW/AAF), not the camera-only index space, and that no driver path uses it (prefer `open_camera_by_id`).
- **odd-NumX/bin-1 comment** — corrected: `rowPitch < (num+1)`-pixel ROI row, so the extra pixel is overwritten by the next row (not a "stride gap"); valid columns stay intact.
- **wait_for_home timeout** — noted the effective wait is up to one poll interval beyond the 6 s budget.

Full sanitizer pre-flight green. Note: the null-model guard is the only behavior change; the rest are comment clarifications.

joeytroy:
Round-22 addressed in 13a007e (both fixed):

- **#1 disconnect vs in-flight exposure** — the connection-task disconnect path now `stop_exposure_thread()` (stop + join) before `set_connected(false)`, matching the destructor, so `Toupcam_Close` can't race a live `Toupcam_WaitImageV4`. The join lives in the connection task, not inside `set_connected` (which already holds `mutex_` → self-deadlock).
- **#2 SetPosition precedence** — a negative position is now rejected as `InvalidValue` before `ensure_connected()`; the upper bound still reports `NotConnected` while disconnected (depends on slot count). Unit test updated.

Full sanitizer pre-flight green.

joeytroy:
Round-23 addressed in 0c72381:

- **#1 get_readout_mode handle/caps race** — real fix: the handle, the CG/HFW capability flags (derived from `camera_info_`), and both SDK register reads now happen under `mutex_` + `readout_mutex_` held together, so they're one consistent connection snapshot (no stale-caps-vs-handle pairing on a disconnect/reconnect). Still atomic against `set_readout_mode`'s two-step apply.
- **#2 rowPitch fragility** — documented the `Toupcam_WaitImageV4` nRowPitch contract at the call site (we pass the unpadded `active_num_x` stride deliberately) with a "revalidate if a future SDK ignores nRowPitch" note.
- **#3 open_camera_by_index** — marked `[[deprecated]]` (no production caller; index space diverges from `enumerate_cameras()` and it opens outside the ref-counted by-id sharing).

Full sanitizer pre-flight green.

joeytroy:
Round-24 addressed in 60ccd46 (both real):

- **#1 AFW double-open leak** — added a `connecting_homing_` flag (under `mutex_`) so a racing synchronous `set_connected(true)` during the homing window short-circuits instead of opening the ref-counted handle a second time.
- **#2 gain/offset stale range bound** — `set_gain` and `set_offset` now read the range, range-check, and write under `mutex_` + `readout_mutex_` held together. The bot's literal "move `offset_max_value()` inside `readout_mutex_`" would invert lock order (that helper takes `mutex_`), so I read `handle_`/bit-depth directly under the held locks. Applied the same fix to `set_gain` too (identical pattern).

Full sanitizer pre-flight green.

joeytroy:
Round-25 — root-cause fix (a1c6516), not another one-method patch:

The recurring stale-handle / use-after-close TOCTOU came from the `handle_copy()` idiom itself — it snapshotted `handle_` under `mutex_` then **released the lock before the SDK call**, so a concurrent `set_connected(false)` could `Toupcam_Close` the handle mid-call. A new `with_handle([&](h){ … })` helper holds `mutex_` **across** the SDK call, and I converted **every** affected getter/setter at once (ccd temp, cooler on/off/power, exposure range, gain get/min/max, IsPulseGuiding, Offset get, cooler set-point get/set, PulseGuide, OffsetMax).

- **The confirmed bug** (`set_readout_mode` stale handle) now holds `mutex_`+`readout_mutex_` and reads `handle_` directly, matching `set_gain`/`set_offset`.
- **Thermal switch**: `elements_` cleared on a failed connect.
- The exposure worker keeps its own snapshot by design (can't hold `mutex_` across `WaitImageV4`; its close is stop-and-joined).
- Lesson captured in AGENTS.md (dcdf84d) so new drivers don't reintroduce it.

Full sanitizer pre-flight green (clean build dir).

joeytroy:
Fixed the round-26 finding in `72ab4b8`: `start_exposure` was the one remaining `handle_copy()` read-then-use — it validated the requested duration via `get_exposure_range(handle_copy())`, releasing `mutex_` before the SDK call. The range read now runs under `with_handle()` (`mutex_` held across `Toupcam_get_ExpTimeRange`). The separate `handle` snapshot captured for the exposure thread is intentionally unchanged — that use is safe because `stop_exposure_thread()` joins the thread before any close. Full `RUN_SANITIZERS=1 ci_preflight.sh` green (ASan+UBSan included).

joeytroy:
Fixed the `set_readout_mode` asymmetry in `1f36141`: it derived the mode spec from a pre-lock `readout_mode_specs()` snapshot, then applied the chosen spec's CG/HFW writes under `mutex_` + `readout_mutex_` — so a disconnect+reconnect to a different model in the gap could pair an old model's spec with the new handle (the getter already avoided this). Split `readout_mode_specs()` into a public wrapper + `readout_mode_specs_locked()`; the setter keeps the pre-lock range check (InvalidValue-before-NotConnected precedence while disconnected), then **re-derives the spec and re-validates the index under both locks**, applying from one connection snapshot. Normal-only falls out as a no-write case. Full `RUN_SANITIZERS=1 ci_preflight.sh` green.

joeytroy:
Fixed the getter mirror in `29efb84`: `get_readout_mode` captured the mode list via `readout_mode_specs()` (pre-lock snapshot), then read the live CG/HFW registers under both locks — a model swap in that window matched the new camera's registers against the old model's list. It now derives the spec via `readout_mode_specs_locked()` **inside the same double-lock** as the register reads, so list + capabilities + registers all come from one connection (the symmetry `set_readout_mode` already relies on). Both directions are now consistent. Full `RUN_SANITIZERS=1 ci_preflight.sh` green.

joeytroy:
Addressed both issues from the last review in `730e0fb`:

**1. XSS (`app.js` 425/1032)** — `loadDevices()`/`loadServerInfo()` now wrap `data.ErrorMessage` in the existing `escapeHtml()` helper. Swept the rest of the file: the other `ErrorMessage` uses go through `textContent`, `alert()`, or `new Error()` (not injection points), and app.js is the only web JS file.

**2. Sync `Connected=false` dropped during FW homing** — added a `pending_disconnect_` flag (guarded by `mutex_`): a disconnect landing in the mutex-released homing window now records intent instead of hitting the `connected==connected_` early-return, and the in-flight connect honors it after homing (closes the freshly-opened handle, stays disconnected) rather than publishing Connected. Cleared per attempt and on the failure path. (Async disconnect is already blocked by `connecting_`; the general disconnect-during-connect cancellation remains tracked in #100.)

**Convention note (per-vendor Switch `get_device_state()`)** — this is pre-existing across all 7 switch drivers (ToupTek ×2, ZWO ×3, Player One, iOptron); the `SwitchDriver` base has no default. Four of them (the GPIO PowerBox variants) read the wrapper directly rather than via the public getters, so consolidating into a base method changes their DeviceState output and needs a fresh ConformU run on each (all merged/validated). Tracked as **#107** rather than folded into this filter-wheel PR.

Full `RUN_SANITIZERS=1 ci_preflight.sh` green (ASan+UBSan included).

joeytroy:
Both issues fixed in `fa694a1`:

**1. `is_guiding` HRESULT check** — was `hr == 0`, conflating `S_FALSE` (0x1, a distinct SUCCESS = "not guiding") with negative error HRESULTs (both returned false), so a transient SDK error cleared the in-flight guide flag mid-pulse. Now `throw_on_error(hr, …)` surfaces real errors and the result is `hr == S_OK`. Swept the wrapper — this was the only `== 0`/`== S_OK` HRESULT judgment; everything else already uses `FAILED`/`throw_on_error`.

**2. Async `disconnect()` dropped during homing** — good catch; my round-29 `pending_disconnect_` fix only covered the sync `set_connected(false)` path. The router routes both the legacy `Connected=false` PUT and the Alpaca v2 `disconnect` endpoint through `disconnect()` → `start_connection_task(false)`, which returned at its `connecting_` guard without recording the request. `start_connection_task(false)` now sets `pending_disconnect_` (under `mutex_`, gated on `connecting_homing_`) when a connect is mid-homing, so both disconnect routes are honored. Lock order stays `connection_mutex_` → `mutex_` (verified no inversion — the destructor and `set_connected` never take them in the other order).

Full `RUN_SANITIZERS=1 ci_preflight.sh` green (ASan+UBSan included).

joeytroy:
All three findings fixed in `2fd5531`, each swept across its full sibling set per the new mandatory symmetry gate (AGENTS.md, `ce15e1a`):

**1. `connecting_homing_` reconnect lockout** — `resolve_wheel_locked()` and `open_filter_wheel_by_id()` now run *inside* the try, so the catch always clears the flag on a failed connect (the catch skips the SDK close when the open itself threw, since `handle` is null there). *Sibling check:* the AFW is the only driver with this guard flag — the camera and thermal switch hold `mutex_` across their whole connect, so no equivalent exists elsewhere.

**2. Sync `set_connected(false)` skipping `stop_exposure_thread()`** — `set_connected(false)` now stops+joins the exposure thread *before* taking `mutex_` (it can't be done under the lock: ToupTek's stop takes `mutex_` internally and the thread takes `mutex_` to publish — self-deadlock). The async path's earlier stop remains a harmless no-op. *Sibling sweep:* **SVBONY and Player One had the same hole on both routes** (their async connection tasks call `set_connected` directly with no prior join) — same hoist applied to both, their `stop_exposure_thread()` is lock-free so it's safe pre-lock. **QHY** dismissed: its disconnect already cancels the exposure before closing, and the wrapper's string-id indirection fails closed rather than touching a freed handle. **ZWO** N/A (no exposure thread).

**3. ZWO `set_position` precedence** — negative check moved before `ensure_connected()`. *Sibling sweep:* **Player One Phoenix had the identical bug** — fixed too. ToupTek was already correct. Unit tests in both suites now assert the disconnected `-1 → InvalidValue` path.

**Bonus (proactive cross-driver sweep, same commit):** the redundant-Connect branch in ToupTek/ZWO/SVBONY/Player One cameras called `reset_exposure_state_locked()` — but the Platform-7 `connect` endpoint calls `connect()` unconditionally, so an idempotent reconnect during an exposure flipped CameraState to Idle mid-frame (letting a concurrent register write hit the live integration) or discarded a just-completed image. Now a pure no-op, matching QHY (the one driver that already had it right). Also ran a full adversarial self-review of the ToupTek diff across all 8 recurring defect classes — HRESULT usage, use-after-close, getter/setter TOCTOU, lock ordering, ref-counting, precedence, XSS all verified clean; the two remaining theoretical items (narrow `connecting_` windows outside homing; paired-axis ROI lost-update) are documented on #100 with rationale.

Full `RUN_SANITIZERS=1 ci_preflight.sh` green (ASan+UBSan included).

joeytroy:
Both findings fixed in `4941a59`, swept across their sibling sets:

**1. AFW startup-window disconnect drop** — you were right that mutex contention makes this hittable; I'd underestimated it. Fixed with a small state machine rather than another gate tweak: a new `inflight_is_connect_` atomic records the in-flight task's direction, both disconnect routes set `pending_disconnect_` whenever a *connect* is in flight (startup window or homing poll), `set_connected(true)` consumes the flag at entry (a newer disconnect supersedes the connect — stays disconnected), the post-homing check consumes it mid-poll, and the connection task re-checks after finishing (running the disconnect if the connect already published). The only remaining exposure is a few instructions before `connecting_.store(false)`, which self-heals on the next connect. The full cancellation-token rework for all drivers stays tracked in #100.

**2. Exposure-thread join/close vs spawn** — the deeper issue behind your finding is that `start_exposure`'s spawn was never serialised against ANY joiner: `stop_exposure`'s join racing the spawn's thread-assignment is UB on `std::thread`, and a spawn slipping between the disconnect's join and `Toupcam_Close` hands the new thread a freed handle. Added a per-driver `exposure_lifecycle_mutex_` (ordered before `mutex_`; the exposure thread itself never takes it, so joins under it can't deadlock): `start_exposure` holds it through the spawn, `stop_exposure` through the join, `set_connected(false)` from the join through the close. ToupTek's async task drops its now-redundant unguarded pre-stop. *Sibling sweep:* the same join-vs-assign UB existed in **SVBONY, Player One, and QHY** (pre-existing, all merged) — same mutex applied to all three; destructors exempt (run after the connection thread joins, no client calls in flight); **ZWO** N/A (no exposure thread).

Full `RUN_SANITIZERS=1 ci_preflight.sh` green (ASan+UBSan included).

joeytroy:
Both findings fixed in `f8d08aa`:

**1. Thermal switch dropped disconnect** — fixed with the identical machinery the AFW got: `inflight_is_connect_` + `pending_disconnect_` recorded on both disconnect routes whenever a connect is in flight, consumed at the connect's entry (newer disconnect supersedes) or by the connection task's post-connect re-check. *Sibling sweep:* the **ToupTek camera** shared the exact same `connecting_`-guard drop — same fix applied, so all three ToupTek devices in this PR now handle the race identically. The remaining pre-existing drivers (ZWO/QHY/SVBONY/Player One families) share this template-level shape; that's the AsyncConnectable rework tracked in **#100** — folding a connection-state-machine change into every merged driver here would outsize this PR.

**2. Silent mode-0 fallback** — `get_readout_mode` now logs a warning with the actual CG/HFW register values when they match no enumerated mode.

Full `RUN_SANITIZERS=1 ci_preflight.sh` green (ASan+UBSan included).

joeytroy:
Fixed in `986b97c`:

**1. `pending_disconnect_` leak → next connect silently no-ops** — correct catch, and it was a flaw in my own machinery: the "self-heals at the next connect" behavior was actually "the next connect silently aborts." All three connection tasks (AFW, thermal switch, camera) now clear the flag under `mutex_` immediately before `connecting_.store(false)`, exactly as you proposed. A disconnect raced out in that instructions-wide window still costs the client a retry of the *disconnect*, but never poisons a future connect.

**2a. `wait_for_home` comment** — you're right: the loop condition rejects `waited == kMaxWaitMs` before sleeping, so the total is exactly 6000 ms (60 × 100 ms). Comment corrected.

**2b. ROI clamp comment ("can fire") — declined, with the math:** `max_w = even_max_w / bin` (floor), so `num ≤ max_w` gives `span = num·bin ≤ even_max_w`. When `span` is odd, `span ≤ even_max_w − 1` (since `even_max_w` is even), so after the `+1` even-rounding `span_w ≤ even_max_w`. Your counter-example (`num·bin = even_max_w − 1` → `span_w = even_max_w`) yields **equality**, and the guard is strict (`span_w > even_max_w`) — it does not fire. The "these clamps never fire" comment is accurate as written.

Full `RUN_SANITIZERS=1 ci_preflight.sh` green (ASan+UBSan included).

---

**Maintainer note (@joeytroy):** per AGENTS.md's merge bar — *no confirmed bugs + ConformU 0/0/0 + all gates green* — this PR now qualifies: all three devices ConformU-clean, every CI gate green, and the last two review rounds have converged from real bugs to an instructions-wide race polish, a comment fix, and one incorrect nit. The reviewer is nondeterministic and re-reviews the full diff each round, so a literal zero-finding run is not a reachable stop condition. Recommend merging after this round's review unless it surfaces a confirmed real bug. Deferred architectural work is tracked: #100 (AsyncConnectable connect/disconnect state machine — would delete all three hand-rolled `pending_disconnect_` implementations), #101 (TSan + stress harness), #107 (Switch DeviceState base).

joeytroy:
Fixed in `363b160`, taking your "single atomic state" suggestion rather than the store swap:

**1. Store-ordering race** — swapping the store order only trades which nanosecond window exists (publish `connecting_` first and a reader can pair it with the *previous* task's stale direction). Collapsed both atomics into one `conn_task_` (`Idle` / `ConnectInFlight` / `DisconnectInFlight`) published in a single store — the half-published pair can no longer be observed, killing the class. Applied to all three drivers including the **camera** (same two-store pattern, not cited in the review but swept per the symmetry gate). Net effect: one atomic *fewer* per driver.

**2. Catch-clear comment** — sharpened as suggested: clearing `pending_disconnect_` on a failed connect is by design (the failure already leaves the device disconnected — the request's desired state — and a stale flag must not abort the user's retry connect at its entry check).

Full `RUN_SANITIZERS=1 ci_preflight.sh` green (ASan+UBSan included).

The merge-bar note from the previous comment stands: ConformU 0/0/0 on all three devices, all gates green, and findings have converged to single-instruction windows and comment wording. #100 remains the tracked home for replacing all of this hand-rolled state with a shared AsyncConnectable base.

joeytroy:
Both findings fixed in `7bdb52f`, and finding 1's fix goes further than the snapshot:

**1. `conn_task_` double-read** — a snapshot alone still leaves the recorder able to set the flag after the task's final clear (the leak we just fixed would return through a different door). Root cause: `Idle` was stored without `connection_mutex_`. The task's **tail** (pending re-check + deferred disconnect + `Idle` store) now runs *under* `connection_mutex_` — the same lock every `start_connection_task` holds — so all `conn_task_` transitions are lock-protected: a racing disconnect either records intent against a genuinely in-flight connect (the tail then **runs the deferred disconnect itself**) or observes `Idle` and spawns a real disconnect task. The previously documented instructions-wide dropped-disconnect window is now *eliminated*, not shrunk. `stop_connection_thread` joins outside `connection_mutex_` (the tail takes it; joining under it would deadlock — verified the destructor and the in-`start_connection_task` join of a finished thread are both safe, the latter only reachable after the tail has released the lock). Applied to all three ToupTek devices (the camera shared the identical pattern).

**2. ROI setter lost-update** — you're right, and I was wrong to dismiss this when my own self-review surfaced it earlier. *Sibling sweep:* the `set_roi_size_locked(num_x, get_num_y())` snapshot-clobber is the template across **all five** camera drivers — fixed in ToupTek, ZWO, SVBONY, Player One, and QHY at once. The shared helpers now take `std::optional<int>` per axis (`nullopt` = leave unchanged, resolved under `mutex_`), chosen over an int sentinel so a client-supplied `-1` still hits the existing InvalidValue rejection; each driver's validation/side-effect logic (effective-ROI alignment, dirty flags, ConformU lenient-setter convention) is preserved verbatim.

Full `RUN_SANITIZERS=1 ci_preflight.sh` green (ASan+UBSan included).

joeytroy:
Both fixed in `f1c9288`:

**1. Deferred-disconnect self-block** — correct catch; this was a regression introduced by the tail redesign one round ago. Two-part fix: the tail now publishes `Idle` **before** running the deferred disconnect (still under `connection_mutex_`, so recorders can't misjudge the state — a racer either recorded before the tail's consume or sees Idle and spawns a real task; the deferred and spawned disconnects are idempotent under `mutex_`). Applied to all three drivers. Additionally the AFW gate gains the `!connected_.load()` conjunct the camera/thermal gates already had — the asymmetry that made only the AFW self-block.

**2. Thermal switch auto-numbering** — added `touptek-thermal-camera-index` to `INDEX_FIELDS` as suggested. *Sibling sweep:* `playerone-switch-camera-index` (the Player One thermal switch, same binding pattern) was missing too — added both.

Full `RUN_SANITIZERS=1 ci_preflight.sh` green (ASan+UBSan included).

joeytroy:
Fixed in `fd13dd8`: `serial_number_.clear()` added to the thermal switch disconnect path, matching the camera driver (the failed-connect catch already cleared it). Checked the AFW for the same gap — its disconnect resets `info_ = {}` whose `id` feeds `get_unique_id()`, so it was already clean. `camera_name_` intentionally left: it only feeds switch descriptions, which are gated behind the connection check, so no stale value is observable. Full `RUN_SANITIZERS=1 ci_preflight.sh` green.

joeytroy:
All three addressed in `d5c6eba`:

**1. Offset support-check TOCTOU** — folded: `ensure_blacklevel_supported()` is now `_locked` and runs inside the SAME `mutex_` hold as the SDK call in all four accessors (`get_offset` and `offset_max_value` check inside their `with_handle` lambda; `set_offset` under its `mutex_`+`readout_mutex_` hold after the handle check; `get_offset_min` under a direct lock). A model swap can no longer slip between the capability check and the register access.

**2. `ReadoutModes` while disconnected — reverted, you're right.** The decisive argument is cross-driver consistency, the project's own rule: ZWO, QHY, SVBONY, and Player One all return a static/cached list while disconnected, so ToupTek throwing NotConnected was the odd one out — and clients do enumerate `ReadoutModes` pre-connect. Disconnected, the list derives from the preloaded model caps (or `Normal`); the current-index getter/setter still require a live connection. Unit test updated.

**3. Homing-cap config point — deferred with rationale:** the 6 s cap follows the INDI toupbase reference connect sequence and no field data shows longer resets; adding a config knob preemptively is speculative surface (project policy is to avoid persisted config without a demonstrated need). If AFW hardware in the field shows longer post-flash resets, a follow-up can raise or expose the cap.

Full `RUN_SANITIZERS=1 ci_preflight.sh` green (ASan+UBSan included).

joeytroy:
Addressed in `4a5c686`:

**1. CG/HFW write order** — fixed as suggested: leaving HFW now disables `OPTION_HIGH_FULLWELL` *before* writing `OPTION_CG`; entering HFW keeps CG-first. Ordered-by-direction with a comment explaining the firmware mutual-exclusivity risk.

**2. Deferred close under `connection_mutex_` — deliberate trade-off, deferred to #100 with a note.** The lock-held ordering is exactly what makes the deferred-disconnect protocol race-free; the stall is bounded (one USB close), rare (requires a disconnect racing a connect AND a third request during the close), and blocks rather than drops. Both alternatives (release-then-recheck, or a detached closer thread) add state or thread-lifecycle risk to save occasional hundreds of ms on an already-exceptional path — that's the AsyncConnectable rework's job, now recorded on #100.

**3. `camera_name_` cleanup symmetry** — tidied on both the failed-connect catch and the disconnect path.

Full `RUN_SANITIZERS=1 ci_preflight.sh` green (ASan+UBSan included).

Issue #107, opened 2026-07-01
Consolidate Switch get_device_state() into SwitchDriver base (remove 7 per-vendor overrides)


joeytroy:
## Problem

AGENTS.md is explicit: **"Do not write a per-vendor `get_device_state()`"** — each device base class should implement it once, inline, by calling that device's own public property getters (guaranteeing DeviceState ↔ GET consistency, which is what ConformU checks).

Unlike `CameraDriver`/`CoverCalibratorDriver`, the `SwitchDriver` base (`AlpacaCore/include/alpacacore/switch_driver.h`) has **no** default `get_device_state()`. As a result **all seven** switch drivers override it:

- `touptek/touptek_thermal_switch_driver.cpp`
- `touptek/touptek_switch_driver.cpp`
- `zwo/zwo_switch_driver.cpp`
- `zwo/zwo_asiair_switch_driver.cpp`
- `zwo/zwo_asiair_plus_switch_driver.cpp`
- `playerone/playerone_switch_driver.cpp`
- `ioptron/ioptron_switch_driver.cpp`

Flagged as a non-blocking convention note in the ToupTek AFW PR (#99) review.

## Two divergent shapes

1. **Getter-based** (playerone, touptek_thermal, zwo_switch): iterate `get_max_switch()` and call `get_switch(id)`/`get_switch_value(id)`/`get_state_change_complete(id)` in a try/catch. These already match the AGENTS.md pattern and would be replaced 1:1 by a base method.
2. **Wrapper-direct** (ioptron, zwo_asiair, zwo_asiair_plus, touptek_switch): read `wrapper_.get_value(i)` and derive "on" from `config_.ports[i].pwm_enabled`, hardcode `StateChangeComplete=true`, and gate on `wrapper_.is_open()`. These **bypass** the public getters — the exact inconsistency AGENTS.md warns against.

## Proposed fix

1. Add an inline `get_device_state()` to `SwitchDriver` (inline in the header — an out-of-line virtual would make the vtable a key function and break the per-vendor static-lib link order, per AGENTS.md). Iterate `get_max_switch()`, call the public getters in a per-id try/catch, append a `TimeStamp` via `device_state_timestamp()`.
2. Remove all seven per-vendor overrides.

## Why not in PR #99

[...]

PR #109, opened 2026-07-02
Consolidate Switch get_device_state() into the SwitchDriver base (7 overrides removed)

joeytroy:
Part of #107 (code complete; the issue stays open for the ConformU V4 hardware re-runs).

## What

- `SwitchDriver` now provides the Platform 7 `get_device_state()` once, **inline in the header** (an out-of-line virtual would make the vtable a key function and break the per-vendor static-lib link order, per AGENTS.md). It iterates `get_max_switch()` and reports `GetSwitchN` / `GetSwitchValueN` / `StateChangeCompleteN` through the **public getters** in a per-id try/catch, appending a `TimeStamp` via `device_state_timestamp()`.
- All **seven** per-vendor overrides removed: `touptek_thermal_switch_driver`, `touptek_switch_driver` (StellaVita), `zwo_switch_driver`, `zwo_asiair_switch_driver`, `zwo_asiair_plus_switch_driver`, `playerone_switch_driver`, `ioptron_switch_driver`.

## Behavior notes

- The three getter-based overrides (Player One, ToupTek thermal, ZWO dew heater) are replaced 1:1.
- The four **wrapper-direct** overrides (iOptron, ASIAIR, ASIAIR Plus RK3568, StellaVita) read `wrapper_.get_value()` and derived "on" locally, bypassing the public getters — the exact DeviceState↔GET desync AGENTS.md warns against. I verified each driver's public `get_switch`/`get_switch_value`/`get_state_change_complete` computes the same values the override did (e.g. iOptron `get_switch` = `wrapper_.get_value(id) != 0`, ASIAIR `v > min` ≡ `v > 0`), so connected DeviceState **values are unchanged**; they gain a `TimeStamp` and are consistent by construction.
- **Disconnected** DeviceState now reports only the `TimeStamp` (getters throw `NotConnected` → omitted), matching the CoverCalibrator base. Six unit tests asserting an empty bag updated.

## Acceptance criteria (#107)

- [x] `SwitchDriver::get_device_state()` added inline to the base header, getter-based, with TimeStamp
- [x] All 7 per-vendor overrides removed
- [x] DeviceState ↔ GET consistency preserved (by construction; wrapper-direct value-equivalence verified per driver)
- [ ] ConformU V4 re-run (0/0/0) on affected hardware: iOptron, ZWO (dew heater + ASIAIR + ASIAIR Plus), Player One, ToupTek thermal — **needs the astro.lan rig with each device attached; tracked on #107**

## Test plan

- `RUN_SANITIZERS=1 scripts/ci_preflight.sh` fully green: clang-format, unicode, build+test (vendors OFF/ON), clang-tidy, cppcheck, ASan+UBSan.
- 15 files changed: +86 −122 (net −36 lines).


joeytroy:
Both fixed in the follow-up commit:

1. **Partial triplet on mid-id throw** — all three values are now computed into locals before any `push_back`, so the per-id push is all-or-nothing. *Sibling sweep (auditable):* checked every other device-base `get_device_state()` (camera, telescope, focuser, rotator, filterwheel, covercalibrator, observingconditions) — all push one property per try via the `add()` lambda, so the grouped-push shape existed only in this new Switch base. N/A elsewhere.

2. **Outer catch scope** — the outer `try` now wraps only the `get_max_switch()` call; loop failures are handled solely by the per-id catch, so the "MaxSwitch itself unavailable" comment can no longer fire for an unrelated escape.

`RUN_SANITIZERS=1 ci_preflight.sh` fully green (clang-format/tidy, cppcheck, both builds, ASan+UBSan).

joeytroy:
Both addressed:

1. **Missing `test_zwo_switch.cpp` coverage** — added the disconnected-DeviceState test mirroring the other six. Good catch on the six-vs-seven mismatch.

2. **`final` guard** — applied. *Sibling sweep:* the same documentation-only enforcement existed on all seven other device-base `get_device_state()` implementations, so `final` is now on every base override (camera, telescope, focuser, rotator, filterwheel, covercalibrator, observingconditions), not just Switch. Verified no vendor override exists anywhere (grep across `src/vendors/` and `include/alpacacore/vendor/`), so nothing can break; the root `AlpacaDriver` virtual remains overridable for the bases themselves.

`RUN_SANITIZERS=1 ci_preflight.sh` fully green.

joeytroy:
CI failure on the r2 push was infrastructure, not code: kernel.org's `/pub/software` tree went 404 (homepage is up, the tarball tree is gone), killing the "Build libgpiod v2 from source" step in the all-vendors and clang-tidy jobs. Pushed a mirror-fallback for the download (mirrorservice.org and mirror.math.princeton.edu both serve the byte-identical 2.2.4 tarball — verified checksums locally); the pinned sha256 is unchanged and still gates every source. No driver code changed in this commit.

Issue #104, opened 2026-07-01
Test-harness: make drivers unit-testable without hardware via a fault-injectable wrapper seam


joeytroy:
## Why

Every driver's logic currently calls the real vendor SDK / protocol wrapper directly (`ToupTekSDKWrapper::instance()`, `Toupcam_*`, serial I/O, etc.), so **nothing about the driver's behavior can be exercised in a unit test without the physical device attached.** That means the highest-risk paths — error/throw handling, resource cleanup, ref-counting, reconnect — are validated only by (a) hardware ConformU runs and (b) code review. As we add more hardware, "works this time on my bench" is not the same as "stays correct every time," and review is the only net under the whole class of throw/cleanup bugs.

## What

Introduce a **mockable seam** at the wrapper boundary so a test can substitute a fake SDK/protocol backend:

- Define a narrow interface (or a set of injectable function objects) for the wrapper operations a driver uses (`open_*`, `close_*`, `put_*`/`get_*`, `start_pull_mode`, `wait_image`, `get_filter_wheel_position`, serial read/write, …).
- Let the driver take the wrapper via that interface (constructor injection / a settable factory) instead of hard-calling the singleton. Production wires up the real wrapper; tests wire up a fake.
- Ship a **fake backend** that can be scripted to: throw from any specific call, return canned enumerations / positions / frames, count opens vs closes, and force error orderings.

## What this would have caught (evidence from PR #99)

- **Connect-path handle leak** (r18 #1): a fake that throws from `put_trigger_mode` / `get_serial_number`, asserting the handle is closed and a subsequent reconnect succeeds (ref count balanced).
- **`close_shared` ref-count underflow** (r19 #3): a test that opens a shared id twice, closes three times, and asserts no close-on-in-use and no underflow.
- **Enumeration filtering / index math**: assert `cameraIndex` skips AFW/AAF entries given a canned mixed enumeration.
- Makes the poll-loop state machines (see companion issue) testable too.

## Scope notes

- Start with ToupTek (most active surface), but the interface pattern should be reusable so ZWO / Player One / QHY / SVBONY and the protocol-wrapper mounts can adopt it — one seam shape, per the "fix shared patterns everywhere" rule.
- This is the **enabler** for the ThreadSanitizer + connect/disconnect/operate stress harness (#101) and complements the config round-trip tests (#102).

Issue #101, opened 2026-07-01
Add a ThreadSanitizer CI job + per-driver connect/disconnect/operate concurrency stress harness


joeytroy:
## Motivation

The #1 driver-review bug class is **concurrency** — use-after-close, TOCTOU on locks, ref-count races, exposure-vs-register-write. But **nothing automated exercises it**: the CI `sanitizers` job is ASan+UBSan (single-threaded memory/UB), and ConformU is single-threaded. So these bugs are caught only by code review, which drove most review rounds across driver PRs (#99, #97, #92, #80, #73).

AGENTS.md already admits this ("ConformU is single-threaded and catches none of the races below — only code review does"). This issue closes that gap with tooling.

## Scope

- Add a `sanitizers-tsan` CI job (or extend `sanitizers`) building with `-fsanitize=thread`.
- Write a **reusable stress harness** that, for a given driver, hammers `connect()`/`disconnect()` from N threads concurrently with operational calls (get/set properties; camera: start/abort exposure, gain/offset/readout-mode; wheel: `set_position`; switch: get/set switch value), asserting no TSan report and no crash.
- Apply first to the drivers with known race history: ToupTek camera + filter wheel + thermal switch, and the ZWO/Player One filter wheels.
- Wire it into `scripts/ci_preflight.sh` behind a `RUN_TSAN=1` knob (mirroring the existing `RUN_SANITIZERS=1`).

## Acceptance criteria

- A data race / use-after-close introduced into a driver fails CI (or the local TSan pre-flight) instead of shipping to code review.
- The harness is easy to extend to a new driver (one small registration).

## References

- Cross-PR retro: 6 of 7 driver PRs with review history shipped at least one concurrency-related fix-induced regression.
- `AGENTS.md` → "Continuous Integration and Pre-flight" (coverage-gap note) and "Driver concurrency & lifecycle".
- Related: #100 (shared base would give these tests one target to exercise).

PR #110, opened 2026-07-02
scan-build evaluation: advisory RUN_SCAN_BUILD pre-flight knob + dead-store fixes

### LINKED ISSUE #106: Evaluate scan-build (Clang Static Analyzer) as an additional CI static-analysis gate
## Why

Our static-analysis gates today are clang-tidy (changed lines) and cppcheck (changed files). Neither does deep **path-sensitive** analysis. Clang Static Analyzer (`scan-build` / the `clang --analyze` engine) does symbolic execution across branches and can flag leak/use-after-free/null-deref paths that line-oriented linters miss — e.g. a handle opened on one path and not released on an exceptional path, or a use-after-close reached only through a specific branch ordering.

This is a **low-to-moderate confidence** addition: the analyzer is genuinely weak across custom RAII, `try/catch` cleanup, and singleton-mediated ownership — which is exactly where several PR #99 bugs lived — so it may not have caught our specific cases. The value is evaluating whether it catches *enough* of the broader leak/null/UAF class to be worth the CI time and the false-positive triage burden as we add more hardware.

## What

1. Add a **non-blocking / experimental** `scan-build` job (or a local `scripts/ci_preflight.sh` opt-in flag, e.g. `RUN_SCAN_BUILD=1`) that runs the analyzer over an all-vendors build and publishes the HTML report as a CI artifact.
2. Triage one full run: count true positives vs noise, note which checkers fire usefully (`unix.Malloc`, `cplusplus.NewDelete`, `core.*`, leak checkers) and which are pure noise on our code.
3. **Decide**: promote to a blocking gate (with a curated checker set + inline suppressions), keep it advisory/artifact-only, or drop it. Record the decision here.

## Acceptance

- One triaged scan-build report attached, with a written keep/advisory/drop recommendation and, if kept, the exact checker set + how suppressions are annotated.

## Relationship to the rest of the test backlog

Complementary to, not a substitute for, the dynamic testing work (#101 TSan + stress harness, #104 fault-injectable wrapper seam, #105 poll-loop tests) — static analysis catches a different (and narrower) slice than fault-injection unit tests do.

joeytroy:
Resolves #106 — the full triage and keep/advisory/drop decision is recorded in the issue comment.

## Decision: keep as ADVISORY local knob, no CI gate

`RUN_SCAN_BUILD=1 scripts/ci_preflight.sh` runs the Clang Static Analyzer over the all-vendors build (mirrors the `RUN_SANITIZERS` pattern). Findings are reported with HTML output under `AlpacaHTTP/build-scan/scan-report/` and never fail the gate; the gate fails only if the analyzer itself can't run.

## Evaluation evidence (clang-19, full all-vendors tree)

18 findings on the pre-fix tree:

- **12× `unix.BlockInCriticalSection`** — every one flags a bounded (`VMIN`/`VTIME` / `SO_RCVTIMEO`) serial or socket read held under the protocol-wrapper mutex: bisque, celestron ×2, gemini, ioptron ×3, synscan ×2, zwo-mount ×2, socket_utils.h. That is the intentional transaction-atomicity architecture (a command/response pair must be atomic under the wrapper mutex). The checker is **disabled in the knob's invocation** — it would re-report the same 12 intentional sites forever.
- **6× `deadcode.DeadStores`** — all real, all trivial, **fixed in this PR**: three never-called copy-pasted `parse_string` lambdas in `router.cpp` (~60 dead lines), a `location_written = true` after its final read, a mid-cycle `last_level = 1` PWM-cache store the unconditional end-of-cycle store made unreadable (ASIAIR Plus RK3568), and `logging_adapter.cpp`'s persisted-level switch now lets the `"INFO"` initializer double as the `Info` case.
- **0 findings** in the leak / use-after-free / null-deref classes that motivated the evaluation — as the issue predicted, the analyzer does not see through the project's RAII + try/catch cleanup and singleton-mediated ownership.

Re-run after fixes + checker curation: **0 advisory findings** (clean baseline going forward).

Why not a CI job: ~10 min of analyzer build per PR to guard a class it demonstrably can't see here, plus triage burden. The dynamic work (#101 TSan, #104 fault-injection) covers that ground properly.

Also picked up by the touched-file gates: `device_path()` in the ASIAIR Plus Impl now returns by const reference (cppcheck `returnByReference`).

## Test plan

- `RUN_SANITIZERS=1 RUN_SCAN_BUILD=1 scripts/ci_preflight.sh` fully green end-to-end (scan-build gate exercised for real: 0 findings).
- All lint gates PASS on the committed diff (clang-format, clang-tidy, cppcheck, shellcheck).


joeytroy:
Round addressed:

1. **Silenced configure diagnostics — valid, fixed.** Configure output goes to `${SCAN_LOG}` (build appends), and the failure branch now labels + tails that log, so a configure failure surfaces its actual error instead of stale output.

2. **"Wrapping cmake configure with scan-build is a no-op" — verified and refuted before applying** (per the AGENTS.md verify-the-reviewer's-fix rule). scan-build interposes by exporting `CC`/`CXX` to the analyzer shims **at configure time**, and cmake caches that compiler. Empirical check on this tree: wrapped configure → `CMAKE_CXX_COMPILER:FILEPATH=/usr/share/clang/scan-build-19/libexec/c++-analyzer`; unwrapped → `/usr/bin/c++`. Applying the suggestion would have kept the gate green while analyzing **zero** translation units — the silent-false-negative failure mode. The wrapper stays, and I've added a `CMakeCache.txt` grep guard that fails the gate loudly if the analyzer compiler is ever not cached, so a future refactor can't reintroduce this silently.

`RUN_SCAN_BUILD=1` preflight re-run end-to-end green (0 advisory findings; shellcheck PASS).

joeytroy:
Verified against the code — the finding is incorrect, so no change is made:

The store removed by this PR is the one in the **post-loop** injection block (`if (in_server_section && !location_written)` *after* the `for` loop). `location_written` is function-local and has **no reads after that point** — the only subsequent code reads `server_section_found`. The in-loop close-of-section injection (`router.cpp:248`) and its `location_written = true` are untouched, as is the in-loop `location:`-key rewrite.

The multi-server scenario in the review runs entirely on the *in-loop* logic: with two `server:` blocks, block 1's close injects and sets the flag, and the post-loop check then skips block 2 — **identically before and after this PR**, because the flag value at the post-loop check comes from the in-loop stores either way. "Location injected into every server block" was never the behavior on either side of this change; the removed store could only have influenced code executing after the function's last read of the flag, of which there is none. The CHANGELOG's "no behavior changes" stands.

Merging per the AGENTS.md merge bar (no confirmed bugs + all gates green + advisory finding refuted with line-level analysis); flagging the disagreement here for the record.

Issue #105, opened 2026-07-01
Deterministic tests for poll-until-settled state machines (injectable clock + input source)


joeytroy:
## Why

Several drivers implement "poll a device until it settles or times out" loops that today (a) call the SDK directly and (b) `sleep` real milliseconds between polls, so they cannot be unit-tested — you'd need the hardware and you'd wait out real time. These loops are exactly where subtle state-machine bugs hide (premature settle, missed timeout, bounce), and they recur across device types as we add hardware.

Examples in-tree:
- ToupTek AFW `wait_for_home` (poll filter-wheel position until homed)
- ToupTek camera exposure watchdog / stuck-exposure deadline
- Mount slew/move-completion polling (iOptron / SynScan / Celestron)

## What

Refactor these loops so the **decision logic is a pure function** over (current reading, elapsed time, prior state), with the **clock and the input reading injected**. Then unit-test the decision logic with scripted sequences — no sleeps, no hardware:

- `wait_for_home`: feed `-1 → -1 → slot → slot → slot` (normal), `slot(bounce) → -1 → slot×3` (deceleration bounce must NOT settle early), fast-homer (`slot×3`, never sees -1), and never-settles (timeout).
- exposure watchdog: feed timestamps that cross / don't cross the deadline.
- slew polling: feed status registers that flip mid-sequence.

## What this would have caught (evidence from PR #99)

- **`wait_for_home` deceleration-bounce false-settle** (r18 #4): a scripted bounce sequence would have failed the test before hardware ever saw it.

## Scope notes

- Builds naturally on the injectable wrapper seam (#104), but the pure-function extraction can also be done standalone.
- Keep the extracted predicate shared/consistent across drivers where the shape matches, so the same test battery guards every poll loop we add for new hardware.

PR #111, opened 2026-07-02
Fault-injectable ToupTek SDK seam + hardware-free driver tests

### LINKED ISSUE #104: Test-harness: make drivers unit-testable without hardware via a fault-injectable wrapper seam

joeytroy:
Resolves #104.

## What

- **`ToupTekSDK` abstract interface** (`touptek_sdk_wrapper.h`): the ~60 SDK operations the drivers use, with the contract documented (throw-on-failure, ref-counted by-id opens). `ToupTekSDKWrapper` implements it (`final`); the deprecated `open_camera_by_index` stays off the interface (no production path).
- **Constructor injection in all four ToupTek drivers** (camera, focuser, filter wheel, thermal switch): the 39 direct `ToupTekSDKWrapper::instance()` call sites now go through the injected `sdk_` member. Default factories pass the singleton — **zero production behavior change** — and each factory gains an overload taking `ToupTekSDK&`.
- **`FakeToupTekSDK`** (`tests/fake_touptek_sdk.h`): scripted backend — throw from any named call (`throw_from.insert("put_trigger_mode")`), canned camera/focuser/wheel enumerations, reference-counted open/close accounting with underflow detection, scripted wheel-position sequences, per-call counters.
- **Five test cases** (`test_touptek_fake_sdk.cpp`, 25 assertions) reproducing the exact PR #99 bug classes that previously only code review could catch:
  1. Connect-path failure (throw from `put_trigger_mode`) must release the ref-counted open, and the same driver reconnects cleanly after (the r18 handle-leak).
  2. `cameraIndex` resolves as a position into the camera-only enumeration, opened by id.
  3. Camera + thermal switch share ONE physical open; disconnecting the camera keeps the switch alive; last holder closes (the r19 ref-count scenario).
  4. AFW homing must not settle on a deceleration bounce (scripted `-1,-1,2,-1,0,0,0` sequence — the r18 false-settle).
  5. Failed thermal-switch connect releases the shared open.
- **AGENTS.md**: seam usage rules — new ToupTek code reaches the SDK only via `sdk_`; connect-path fixes come with a fake-SDK repro test; the seam shape (interface + factory overload + scripted fake) is the reusable pattern for other vendors when their wrappers are next touched.

This is the enabler for #101 (TSan stress harness gets a hardware-free backend to hammer) and #105 (poll-loop tests).

## Test plan

- All five fake-SDK tests pass; full suite green.
- `RUN_SANITIZERS=1 scripts/ci_preflight.sh` fully green (clang-format/tidy, cppcheck, both builds, ASan+UBSan).
- Production wiring unchanged: default factories forward to `ToupTekSDKWrapper::instance()`; router code untouched.


joeytroy:
All three addressed: the two `override`s the sweep script missed (both declarations carry trailing comments, which the line-end match skipped — swept the header again, these two were the only such lines), the focuser connect-path failure test (throw from `aaf_range` during MaxStep discovery; ref count balanced + clean reconnect — completes the four-driver set), and `close_shared` restructured to an iterator-based loop with no statement between `erase` and `return`. Six fake-SDK tests now, 33 assertions. `RUN_SANITIZERS=1 ci_preflight.sh` fully green.

joeytroy:
Round 3 addressed (all in the fake, as noted):

1. **`aaf_range` per-action results** — now honors `(range_action, target_action)` via a scriptable `aaf_range_script` map (default 100000), so per-action range validation is testable.
2. **`ref_counts_` map hygiene** — the entry is erased on physical close; the two maps can no longer diverge across open/close cycles.
3. **`get_size` per-handle** — resolves the handle to *its* camera's dimensions; multi-camera tests with differing sensors now read correctly.
4. **Taillight default — kept ON, with the rationale documented in the fake**: the real ATR2600M cold-start lights the tail LED, which is precisely why the thermal switch exposes the element (users turn it *off* to avoid reflections/light leaks — per the AGENTS.md ToupTek notes). A fake that boots dark would misrepresent the hardware and hide the driver's most common first action. Documented rather than changed.
5. **Lifetime rule** — stated prominently in the `FakeToupTekSDK` class comment (fake must outlive every driver constructed on it; declare fake first). Kept as documentation rather than pointer+destructor-assert: the drivers' production contract is a singleton that always outlives them, and adding assert machinery to production drivers for a test-only hazard inverts the trade-off.

`RUN_SANITIZERS=1 ci_preflight.sh` fully green; 6 fake-SDK tests, 33 assertions.

joeytroy:
Fixed: the fake canonicalizes all eight AAF Set→Get action pairs (Position, Direction, MaxIncrement, Fine, Coarse, Buzzer, Backlash, MaxStep) so writes round-trip on reads, plus a position round-trip test (`move(1234)` → `get_position() == 1234`). Seven fake-SDK tests, 34 assertions; `ci_preflight.sh` fully green.

joeytroy:
Round 5 addressed:

1. **`get_final_size` decoupled** — own body via a shared private `lookup_size`; each SDK call is now independently faultable and counted, matching the real SDK's distinct `get_Size`/`get_FinalSize`.

2. **Shared handle namespace — intentional, refuting with the contract**: `Toupcam_Open` has ONE id namespace across cameras, focusers, and wheels; that's the entire basis of the ref-counted shared open (AGENTS.md ToupTek notes: *"two driver instances on the same id (the camera + its thermal switch, or a camera + its integrated autofocuser) now correctly share one open"* — and *"ALL open-by-id paths route through these helpers — camera, focuser, AND filter wheel"*). A camera and focuser sharing an id is not a test accident to detect — it's the integrated-autofocuser hardware reality the fake must model. Adding a type discriminator would make the fake *diverge* from the wrapper it stands in for. Kept as-is.

3. **Pre-existing focuser race — fixed here rather than deferred** (project rule: a template bug found in one driver gets fixed immediately, and this PR is the one that makes the file's SDK calls seam-visible). `get_is_moving`/`get_position`/`get_temperature`/`halt`/`move` now hold `mutex_` across the SDK call (checklist shape (a)), so the connected check and the `aaf_*` call can't straddle `set_connected(false)`'s close. `move` keeps a single hold across validation and the write; exception ordering unchanged. *Sibling sweep:* every other `handle_` use in the focuser is inside `set_connected` (already under `mutex_`); the camera uses `with_handle`, and the wheel/thermal drivers take locked reads — no other unguarded sites.

`RUN_SANITIZERS=1 ci_preflight.sh` fully green; full suite 325 assertions passing.

joeytroy:
Round 6 addressed:

1. **`get_max_step`/`get_max_increment`** — lock now precedes the connected check, matching r5's five methods. *Sibling sweep (vendor-wide, auditable):* grepping every `ensure_connected(); → lock_guard` adjacency across the four ToupTek drivers found **four more camera sites** with the same shape — `get_image_array`, `get_offset_min`, `set_roi_size_locked`, `set_start_pos_locked` — all reordered identically in this commit. The thermal switch's two adjacencies already re-assert `ensure_connected_locked()` under the lock (the documented PR #99 pattern) and keep ASCOM's validate-before-connect ordering, so they're correct as-is.

2. **`close_shared` map access** — `find()` instead of `operator[]`, so a future divergence of the two maps corrupts nothing silently.

`RUN_SANITIZERS=1 ci_preflight.sh` fully green (both build configs, ASan+UBSan).

PR #112, opened 2026-07-02
Deterministic poll-until-settled decision logic (util::ConsecutiveSettle) + scripted tests

### LINKED ISSUE #105: Deterministic tests for poll-until-settled state machines (injectable clock + input source)

joeytroy:
Resolves #105.

## What

- **`util::ConsecutiveSettle`** (`util/poll_settle.h`): the "poll until settled or timed out" *decision* — where premature-settle, missed-timeout, and final-poll off-by-ones live — extracted into a pure state machine: consecutive-stable-reads requirement with bounce reset, poll budget, settle-beats-timeout on the final poll, sticky terminal states. No clock needed: one feed per poll period makes the poll count the elapsed time, so tests are pure sequence scripts.
- **ToupTek AFW `wait_for_home` rewired onto it** with byte-identical semantics (3 stable reads, 60×100 ms budget — the loop keeps only the sleep cadence and the SDK read). The #104 fake-SDK integration test still passes through the real driver connect path, proving no behavior change.
- **Eight scripted-sequence tests** (`test_poll_settle.cpp`, 195 assertions, zero sleeps): normal home, deceleration bounce (the PR #99 r18 false-settle, as its own regression test), double bounce, fast homer, timeout exactly at the poll budget, settle-on-final-poll, almost-run that ends moving, sticky terminals.
- **AGENTS.md**: rule for when to use the shared policy — any new stability-run + poll-budget loop; single-edge polls ("until IsSlewing flips") don't need it.

## Scope notes (from the issue's examples)

- **Camera exposure watchdog**: a single deadline compare (`now >= exposure_deadline_`), not a state machine — extracting it would add indirection with nothing to test. Left inline.
- **Mount slew polling** (iOptron/SynScan/Celestron): single-edge completion polls with no stability requirement — no matching shape today. The AGENTS.md rule directs any future stability-shaped loop to the shared policy.

## Test plan

- `[settle]` suite: 195 assertions passing in microseconds; `[fakesdk]` AFW test still green through the rewired driver.
- `RUN_SANITIZERS=1 scripts/ci_preflight.sh` fully green.


joeytroy:
CI failures were infrastructure again — kernel.org has now *purged* old libgpiod point releases entirely and the mirrors synced the deletion, so this morning's mirror-fallback chain 404s end to end. Fixed durably: the pristine 2.2.4 tarball is now **vendored** at `AlpacaCore/external/libgpiod/` (retrieved from the Fedora lookaside cache, verified byte-identical to the kernel.org artifact against the pinned sha256 — provenance in the README next to it). Both CI jobs copy from the repo and still verify the checksum; no network fetch remains. No driver code in this commit.

joeytroy:
Both addressed: `*.xz` (plus `*.gz`/`*.bz2`) marked `binary` in .gitattributes so the `eol=lf` catchall can't corrupt the vendored tarball on Windows checkouts, and the libgpiod version + sha256 pins are hoisted to a single workflow-level `env:` block consumed by both jobs — a bump can't skew them anymore. Preflight green (zizmor included, workflow changed).

joeytroy:
Both addressed:

1. **Construction guard** — `std::invalid_argument` for `stable_reads_required < 1` or `max_polls < 1`, with a test. (Not `assert()`: Release builds define `NDEBUG`, which would compile the guard out exactly where a config-sourced zero would arrive.)

2. **iOptron slew-settle loop — you're right, and my PR text was wrong**: it does share the consecutive-run core; "mount loops are single-edge" is corrected in AGENTS.md and the CHANGELOG. It stays hand-rolled **by design**: its budget is a clock deadline (not a poll count) and it has a dual exit (target-reached OR stabilized-without-target), so adopting the poll-budget API would mean faking a `max_polls` from a sleep interval — a lie waiting for the interval to change. AGENTS.md now names it as the one remaining hand-rolled instance, to be unified if a third loop of that extended shape ever appears.

Preflight fully green; `[settle]` suite now 9 cases including the guard.

joeytroy:
All three addressed: (1) construction now also rejects `stable_reads_required > max_polls` (statically unreachable Settled → silent perpetual TimedOut), with a test; (2) the README and both ci.yml step comments now state the honest scope of the in-repo checksum — corruption, not adversarial modification, with directory changes to be reviewed like a dependency bump against the documented provenance chain; (3) the version-bump instruction points at the real single `env:` block instead of the removed per-job `ver=` pins. Preflight fully green.

joeytroy:
Added the production-boundary test exactly as suggested (option 2): `ConsecutiveSettle(3, 60)` fed `[58×false, true, true]` — poll 60 is a settled read with a 2/3 run and must TimedOut. The (3, 61) variant stays for the reset-on-final-poll shape. `[settle]` suite now 10 cases; preflight fully green.

PR #114, opened 2026-07-02
Config round-trip tests for every driver + two mountIndex data-loss fixes

### LINKED ISSUE #102: Add config save→load round-trip tests for every driver's persisted fields
## Motivation

Two recurring **silent-data-loss** classes across driver PRs, both invisible until a user re-saves a device:

1. **`sanitize_device_config` (router.cpp) drops an un-allowlisted persisted field.** The sanitizer is a strict allowlist; a field the registration path reads but the sanitizer doesn't `copy_if_present` is silently discarded on save. (#99 stripped `filterwheelIndex` / `filterwheelId` / `filterNames`.)
2. **Web-UI `FormData` name collision.** A non-vendor-prefixed form-field `name` collides with ZWO's bare name (ZWO submits even while hidden → its value wins). (#99: ToupTek `focuserId` / `filterwheelId` vs ZWO.)

## Scope

- Extend `AlpacaHTTP/tests/test_routing.cpp`: for each driver's config, POST `configuredevice`, then GET `configureddevices` and **assert every persisted field survives** the round-trip (index/id, `filterNames`, PWM `ports`, `gpioChip`, `switchType`, etc.). Model on the existing ToupTek AFW filter-wheel round-trip test added in #99.
- Cover all vendors/device types that persist config.

## Acceptance criteria

- Adding a config field that a driver reads but isn't allowlisted in `sanitize_device_config` fails a test.
- A FormData field-name collision is catchable by the same round-trip assertions.

## Notes

Now a **required** test per driver (`AGENTS.md` → Testing Requirements → Required Test Cases #6). This issue back-fills the existing drivers.

## References

- PR #99, `AGENTS.md` → Required Test Cases, "Enumeration index fields".

joeytroy:
Resolves #102.

## What

**Seventeen new field-survival round-trip tests** in `AlpacaHTTP/tests/test_routing.cpp` — every (vendor, deviceType) that persists config now POSTs distinctive values through `configuredevice`, reads `configureddevices` back, and asserts **every persisted field survives** `sanitize_device_config`:

- **ZWO**: camera, filterwheel, focuser, rotator, switch × 3 shapes (dew heater; ASIAIR libgpiod with `gpioChip`/`pwmFrequencyHz`/per-port `gpio`+`name`+`pwm`; ASIAIR Plus RK3568 with `devicePath`, plus a negative assert that `gpioChip` doesn't leak in)
- **QHY** (string `cameraId`), **SVBONY**, **Player One** camera (upgraded from configure-only) + filterwheel
- **ToupTek StellaVita** (upgraded from a no-crash test to field survival)
- **Gemini** focuser, **WeeWX** observing conditions
- **iOptron / SynScan / Bisque** telescopes

Shared `roundtrip_config` / `remove_device` helpers keep each block to its distinctive values + assertions; all blocks are `#ifdef`-guarded to match their registration branches (vendor-OFF builds unaffected).

## Two real bugs found by the field inventory, fixed here

`register_device_from_config` reads `mountIndex` for **iOptron** and **SynScan** telescopes (auto-detect device selection), but neither sanitizer branch allowlisted it — a saved non-zero mount index **silently reverted to 0** on the next load. Celestron was the only mount vendor with the key allowlisted. Both branches now `copy_if_present("mountIndex")`, and the new telescope round-trips assert it (they fail without the fix). This is precisely the silent-data-loss class the issue exists to catch.

## Test plan

- `test_routing` passes with all 17 new blocks on the all-vendors build.
- `scripts/ci_preflight.sh` fully green.


joeytroy:
Two updates:

**The CI segfault was a real production bug, now fixed in this PR.** Reproduced locally by hiding USB (`bwrap --tmpfs /dev/bus/usb --tmpfs /sys/bus/usb`), with a clean backtrace: libqhyccd's **first entry point** spawns its `PnpEventListenerThread`, which calls `libusb_hotplug_register_callback` on the failed libusb context → SIGSEGV. So *configuring* a QHY camera (construction-time model preload) or answering a `configureddevices` poll (SDK-version query) crashed the whole server on any USB-less host. Fixes: the wrapper makes **no** libqhyccd call until the first real camera operation (lazy `ensure_resource`, i.e. connect); `get_sdk_version` is served from a cache filled at init (empty → `nullopt` pre-connect); the construction-time preload is removed (web UI shows the model after first Connect — the cost of not crashing the server). `test_routing` now passes in the USB-hidden sandbox where it previously segfaulted at the QHY block — exactly the class of find these round-trip tests exist for.

**Review finding fixed** — `roundtrip_config` guards the `"Value"` subscript, so an error envelope produces a clean `EXPECT` diagnostic instead of an uncaught `json::out_of_range`.

`RUN_SANITIZERS=1 ci_preflight.sh` fully green.

joeytroy:
Both fixed: `get_driver_info` omits the SDK suffix while the cache is empty (no more `(SDK )` pre-connect — matches the `get_device_sdk_version` nullopt behavior), and `roundtrip_config` uses non-throwing `json::parse` with `is_discarded()` checks on both responses. Verified in both the normal and USB-hidden sandbox runs; preflight green.

joeytroy:
Both fixed: `remove_device` now parses the response and `EXPECT`s `ErrorNumber == 0` (cleanup failures fail at the cleanup step), and all seventeen blocks require `cfg.is_object() && !cfg.empty()` so a strip-everything sanitizer regression is caught before the per-field defaults can mask it. Verified in normal + USB-hidden sandbox runs; preflight green.

PR #115, opened 2026-07-05
Complete #100: AsyncConnectable everywhere + #101 TSan stress harness (hardware-validated)

### LINKED ISSUE #100: Extract a shared AsyncConnectable driver base + sweep the shutting_down_ teardown guard across all vendor drivers
## Motivation

The async connection-thread lifecycle (`start_connection_task(bool)` / `stop_connection_thread()` / `connection_thread_` / `connecting_`) is **copy-pasted across ~26 drivers**. Without a `shutting_down_` guard, a `connect()` racing the destructor can spawn a `connection_thread_` that outlives the object and is never joined → `std::terminate`.

Only the **4 ToupTek drivers** (camera / filter-wheel / focuser / thermal-switch) currently have the guard (added in #99). The other **~22 vendor drivers still need it.**

Thread-lifecycle / use-after-close races were the **#1 recurring finding across every driver PR with review history** (#72, #73, #80, #92, #97, #99). Fixing it per-driver is mechanical; extracting a shared base fixes it once and makes every *future* driver correct by construction.

## Scope

- Design an `AsyncConnectable` mixin/base encapsulating the full contract:
  - `std::atomic<bool> connecting_`, `std::mutex connection_mutex_`, `std::thread connection_thread_`, `bool shutting_down_` (under `connection_mutex_`);
  - `start_connection_task(bool)` that returns early when `shutting_down_`;
  - `stop_connection_thread()`; and the destructor teardown contract (set `shutting_down_` under the lock **before** join).
- Migrate all vendor drivers: ZWO (camera/telescope/focuser/filterwheel/rotator/switch + asiair/asiair-plus switch), Player One (camera/filterwheel/switch), QHY, SVBONY, iOptron (switch/telescope), SynScan, Celestron, Bisque, Gemini, WandererAstro, WeeWX, ToupTek StellaVita switch.
- **Note:** the pattern is not perfectly uniform — synscan / celestron / bisque / ioptron_telescope use an atomic-only variant *without* `connection_mutex_`. The base must accommodate them, or those get the guard inline.
- Also fold in the "join **every** background thread in the destructor" rule (exposure/pulse-guide timers), and consider hosting the lock-order discipline (`mutex_` → op-lock → SDK) as documented invariants of the base.

## Acceptance criteria

- No driver can spawn a connection thread that outlives the object.
- Ideally one shared implementation; each driver's unit tests + ConformU still pass.
- AGENTS.md "Driver concurrency & lifecycle" updated to reference the base (remove the "copy-pasted across ~26 drivers … follow-up" caveat).

## References

- PR #99 (the guard was added to the 4 ToupTek drivers there)
- `AGENTS.md` → "Driver concurrency & lifecycle (read before writing a driver)"

### LINKED ISSUE #101: Add a ThreadSanitizer CI job + per-driver connect/disconnect/operate concurrency stress harness

joeytroy:
## Summary

Four commits closing the two concurrency follow-ups from the PR #99 retro, hardware-validated end to end:

1. **`AsyncConnectable` base + reference migration** (f2359c6, already on this branch): the battle-tested PR #99 connection-thread protocol extracted once.
2. **Migrate all 25 remaining vendor drivers** (#100): every one of the 26 vendor drivers now inherits the shared mixin — no driver carries its own `connection_thread_`/`connecting_` machinery. The 22 non-ToupTek drivers gain both the `shutting_down_` destructor guard (connect racing `~Driver` → unjoined thread → `std::terminate`) and the never-drop-a-racing-disconnect protocol they never had. Base hardening: a throwing `std::thread` ctor rolls back the in-flight publish; new `record_pending_disconnect()` hook for the AFW's sync-connect homing window.
3. **TSan CI job + concurrency stress harness** (#101): reusable `[stress]` suite (lifecycle storm / destruction-during-connect / racing-disconnect-settle) registered for ToupTek camera/AFW/thermal (fake-SDK seam via a new thread-safe `LockedToupTekSDK` decorator), ZWO EFW, Player One Phoenix. New `sanitizers-tsan` job runs it under ThreadSanitizer, all vendors ON; `RUN_TSAN=1` pre-flight knob mirrors `RUN_SANITIZERS`.
4. **ConformU 4.4.0 re-validation sweep (#107) + RK3568 `.deb` fixes + thermal-first ToupTek UI.**

## Hardware validation

Eight clean ConformU 4.4.0 runs (0 errors / 0 issues / 0 timing) on builds from this branch, across six machines, covering seven migrated drivers:

| Device | Machine | Notes |
|---|---|---|
| ZWO dew heater (ASI2600MM Pro) | RPi rig | #107 ✔ |
| ToupTek thermal switch (ATR2600M) | RPi rig | #107 ✔ wrapper-direct |
| iOptron iMate PowerBox | iMate (Armbian/mainline) | #107 ✔ wrapper-direct |
| ToupTek StellaVita PowerBox | StellaVita CM4 | branch validation |
| Player One thermal (Uranus-C PRO) | StellaVita CM4 | #107 ✔ |
| ZWO ASIAIR Pro | Pi 4 | shared `asiair` driver |
| ZWO ASIAIR Plus (Pi CM4) | CM4 | #107 ✔ wrapper-direct |
| ZWO ASIAIR Plus (RK3568) | stock ZWO kernel 4.19.219 | #107 ✔ + packaging fixes proven |

All four wrapper-direct DeviceState drivers from PR #109 confirmed on the getter path. Issue #107's checklist is fully ticked (details on the issue).

## Variant notes for review

- Wrapper-state switches (StellaVita / iMate): no driver mutex; the base gates read `wrapper_.is_open()` as connected-now.
- Gemini + WandererAstro keep their documented synchronous `disconnect()` (base `stop_connection_thread()` + sync `set_connected(false)`).
- iOptron telescope's old `if (connect == get_connected()) return;` spawn-skip is dropped: the base spawns and the task's idempotency check no-ops — same observable behavior.
- ToupTek AFW keeps `connecting_homing_` (sync-connect double-open guard the base's `conn_task_` gate can't see); its pending-disconnect record/consume now uses the base flag via `record_pending_disconnect()`.

## Gates

- `scripts/ci_preflight.sh`: all mandatory checks PASS (clang-format, unicode, build+test vendors OFF/ON, clang-tidy, cppcheck); zizmor + shellcheck run manually on the changed workflow/scripts — clean.
- `alpacacore_tests`: 238 cases + 13 new `[stress]` cases pass; the stress suite also passes under a full all-vendors TSan build with **zero** ThreadSanitizer warnings.

## Out of scope (follow-ups)

- 14 pre-existing `.detach()` background timers across 7 vendors (pulse-guide/goto watchers, some capturing `this`) — separate bug class, needs its own issue.
- Extending `[stress]` registration to more drivers as fake seams appear.

Closes #100. Closes #101. (#107 can close once this merges — all six re-validations are recorded on the issue.)


joeytroy:
Both findings addressed in c524652 (plus the clang-format CI failure):

1. **catch(...) hardening** — added the fallback in both thread-body sites (primary `set_connected` call and the deferred disconnect). Sibling sweep: the `start_connection_task` ctor-rollback already used `catch (...)`; the per-driver destructor `catch (const std::exception&)` teardown guards are pre-existing, per-driver, and outside this base — unchanged.
2. **Deferred-disconnect latency** — now documented explicitly in the class comment as a deliberate COST note (correctness over latency; only on the disconnect-races-connect path; flagged for future drivers with slow teardowns).

clang-format: expanded the over-limit one-liners in `locked_touptek_sdk.h`/`concurrency_stress.h` via `git-clang-format` against the merge-base — local pre-flight had missed them because both files were untracked when the changed-lines check ran.

Re-validated: 251 unit tests pass; `[stress]` passes under the all-vendors TSan build with zero warnings.

joeytroy:
Round-2 finding fixed in b595aef — and it was a good catch, thank you.

**ABBA deadlock**: applied the suggested hoist (verified before applying: the `get_connected()` read was always a momentary snapshot — the sync setter never held `mutex_` and `pending_mutex_` around it simultaneously — and a stale read in either direction is benign since the deferred `set_connected(false)` is idempotent; flag-consumption order races pre-exist with identical settled outcomes). `get_connected()` now runs under `connection_mutex_` only, consistent with the documented lock order.

**Coverage gap**: also addressed — new `test_bisque_concurrency_stress.cpp` adds the mutex-guarded-`get_connected()` driver shape to the `[stress]` pool (hardware-free via instant `ECONNREFUSED` on a loopback port). Verified the coverage actually bites: with the fix reverted, TSan reports `lock-order-inversion (potential deadlock)` on exactly this test; with the fix, zero warnings. The `sanitizers-tsan` job would now catch a regression of this bug class for all four telescope drivers sharing the shape.

Validated: 253 unit tests, 15 `[stress]` cases under all-vendors TSan (zero warnings), cppcheck + clang-format clean.

joeytroy:
Round-3 findings addressed in fcc7baf:

1. **ZWO dew-heater close ordering** — fixed by reordering to clear-state → `connected_.store(false)` → SDK close, mirroring the EFW/CAA/EAF siblings. The `camera_id_` non-reset in the explicit-id branch is intentional (it is the durable config used by the next connect), so only the ordering changed. **Sibling sweep:** the Player One thermal switch had the identical shape (snapshot-then-call ops + close-first order) and is fixed in the same commit. The ZWO/Player One/SVBONY *camera* drivers share the ordering but their disconnects interact with exposure-worker lifecycles where a blind reorder risks regressions — audited and split to #116 with per-driver evidence (ToupTek camera via `with_handle`+lifecycle-mutex and QHY via mutex-held ops confirmed safe and exempt).
2. **`settle_connected` fallback** — now requires `!get_connecting()` like the loop exit, so a wedged driver can't vacuously pass a `want=false` check.
3. **Connect-during-disconnect asymmetry** — documented as deliberate in the class comment with the safety rationale: a dropped disconnect leaves hardware energized unattended (invisible to the user); a dropped connect is immediately visible (`Connected` stays false) and trivially retried by the client. Mirroring `pending_connect_` would double the tail/entry protocol surface for a case with a benign failure mode.

Validated: 253 unit tests; `[stress]` under all-vendors TSan with zero warnings; clang-format clean.

joeytroy:
Round-4 findings addressed in f061ed5:

1. **Disconnect-loss via no-op connect** — fixed in the base rather than either suggested option, because both have a gap: (a) a spawn-skip guard only covers the async path — the same loss exists when a client calls the *sync* `set_connected(true)` on a connected device while a recorded disconnect is pending; (b) reordering idempotency-before-consume per driver is 26 hand edits with no compiler check. Instead `consume_pending_disconnect(bool connected_now)` now consumes only when a REAL connect is about to happen; a no-op connect (async or sync) leaves the flag for the task tail's deferred disconnect. The signature change makes the sweep compiler-enforced — all 29 call sites updated with each driver's record-gate expression; the AFW's three cleanup consumes run in disconnected windows so their semantics are unchanged. New deterministic harness check `connected_then_connect_disconnect_settles_disconnected()` encodes your exact scenario and is wired into the ToupTek stress suite.

2. **Record TOCTOU** — closed structurally rather than tracked: the tail now publishes Idle *before* its pending-consume, and the recorder re-reads `conn_task_` *inside* `pending_mutex_`. The mutex serializes recorder vs tail-consume: recorder-first ⇒ tail is guaranteed to see the flag; tail-first ⇒ Idle was published before the tail released the lock, so the recorder observes Idle, records nothing, and the caller performs the sync disconnect itself. No stale-flag window remains, and the Idle-before-deferred-disconnect property the sync gate depends on is preserved.

Validated: 254 unit tests; 16 `[stress]` cases under all-vendors TSan with zero warnings; clang-format clean.

joeytroy:
Round-5 items addressed in the latest push:

1. **QHY dead gates** — each branch of the split `set_connected_impl` now carries only its relevant obligation (4 disconnect-side, 5 connect-side), matching the AFW/ZWO split-branch style. The dead checks were an artifact of the round-4 compiler-enforced sweep being applied uniformly; statically false, no behavior change.
2. **Suppressions scope** — documented the residual risk precisely: `called_from_lib` swallows driver-code races on SDK-callback stacks, symbol-level `race:` entries aren't possible against the unsymbolized blobs, and callback-path coverage belongs in the fake-SDK-seam stress tests where no suppression applies.

Noted the review's "neither issue requires blocking the merge" — both fixed anyway to land clean. 254 unit tests pass, clang-format clean.

joeytroy:
Round-6 finding fixed in the latest push — the suggested fix applied verbatim after verification (record independent of `connected_now`, return skip-hardware only when `!connected_now`). Traced all four interleavings clean: sync-disconnect-while-connected now leaves the flag, the no-op task consumes it at entry and stays down, the tail sees no flag with `connected_now=false` so no double-teardown, and the round-4 TOCTOU pairing (conn_task_ re-read under `pending_mutex_` vs Idle-before-consume) is unchanged.

Test gap closed as suggested: the harness check gained a `use_sync_disconnect` variant and a new ToupTek stress case exercises the sync PUT path. Negative-tested: with the round-4 recorder temporarily restored, the new test fails exactly as described (no-op task reconnects); with the fix, clean.

255 unit tests; 17 `[stress]` cases under all-vendors TSan, zero warnings.

joeytroy:
Round-7 items addressed in the latest push:

1. **iOptron deferred-disconnect latency** — documented at the driver's `connect()/disconnect()` overrides as the worst case of the base's COST note (teardown joins the clock-sync thread through multi-second serial retries while `connection_mutex_` is held), along with the always-spawn rationale from the minor note (the old spawn-skip masked the round-4 flag-consumption bug).
2. **Zero-matching-tests green-light** — both the CI job and the `RUN_TSAN=1` pre-flight now require ≥1 discoverable `[stress]` test (`--list-tests --verbosity quiet`, the Catch2 v3 spelling) before running the suite; 17 currently discovered.
3. **Callback-path coverage** — AGENTS.md's `[stress]` registration rule now states explicitly that a new SDK callback must ship with a fake-seam stress path in the same change, since green TSan CI cannot see races on vendor-blob stacks.

255 unit tests pass; shellcheck/zizmor/YAML/clang-format clean.

PR #119, opened 2026-07-06
Camera close-order audit + held-mutex conversion, exposure-failure recovery, TSan guard fix

### LINKED ISSUE #116: Audit camera drivers: SDK close ordered before connected_ clear in set_connected(false) (same class as PR #115 round-3 switch finding)
PR #115's round-3 review found the ZWO dew-heater switch closing the SDK camera **before** `connected_.store(false)`, letting an operational call that passed `ensure_connected()` race onto a just-closed handle (its ops snapshot the id and call the SDK without holding `mutex_`). That instance and its exact template sibling (Player One thermal switch) were fixed in the PR by reordering per the AGENTS.md disconnect rule (clear state → publish disconnected → close last), matching the ZWO EFW/CAA/EAF pattern.

The sweep found the same close-before-clear **ordering** in the camera drivers, which were deliberately NOT blanket-reordered in a review round because their disconnect paths are entangled with exposure-worker lifecycles (a reorder can change what an in-flight exposure thread observes):

- **ZWO camera** (`zwo_camera_driver.cpp`): `stop_exposure` + `close_camera` before `connected_.store(false)`; ops shape needs confirmation.
- **Player One camera** (`playerone_camera_driver.cpp`): `close_camera` before store; ops are snapshot-then-call (`camera_id_copy()` then SDK without `mutex_`) — likely the same exploitable window as the switches.
- **SVBONY camera** (`svbony_camera_driver.cpp`): `stop_video_capture` + `close_camera` before store; ops shape needs confirmation.

Confirmed NOT affected (safe shapes):
- **ToupTek camera**: `with_handle` holds `mutex_` across SDK calls; exposure teardown serialized by `exposure_lifecycle_mutex_` (PR #99 review artifact).
- **QHY camera**: ops hold `mutex_` across the SDK call.
- **ToupTek focuser / Player One filterwheel**: ops hold `mutex_` and re-check under the lock.

Work: for each of the three, map the exposure-thread interactions with `connected_`/camera state, then either reorder (store-then-close) or convert ops to the held-mutex shape — one PR, with `[stress]` registrations added where a fake seam exists. Per the fix-shared-patterns-everywhere rule this must cover all three in one change.

### LINKED ISSUE #117: ci_preflight.sh TSan zero-test guard is Catch2 v3-only — false PASS possible under Catch2 v2
Follow-up from PR #115 round-8 review (bot approval noted this as low-severity, non-blocking).

## Problem

The `RUN_TSAN=1` block in `scripts/ci_preflight.sh` guards against silently running zero stress tests by counting the output of:

```
./AlpacaCore/build-tsan/tests/alpacacore_tests --list-tests --verbosity quiet "[stress]" | wc -l
```

`--list-tests --verbosity quiet` is Catch2 **v3** syntax. Under Catch2 **v2**, a tag filter matching zero tests still prints one line (`0 matching test cases`), so the `wc -l >= 1` check always passes — a developer running local preflight with v2 installed could get a PASS while zero `[stress]` tests actually executed.

**Scope:** local preflight only. GitHub CI is unaffected (Ubuntu 24.04 ships Catch2 v3.4.0), and the project builds against v3, so a v2 environment is unlikely to build the tests at all. The `test -x` binary-existence guard still catches the silent-cmake-skip case.

The same pattern exists in the `sanitizers-tsan` job in `.github/workflows/ci.yml`; it is safe there (runner always has v3) but should be kept in sync with whatever fix lands here.

## Suggested fix

Make the guard version-agnostic by asserting on what actually *ran* instead of what the list command printed — parse the Catch2 run summary (`test cases: N | ...`) after executing the `[stress]` suite and require N ≥ 1. Alternatively (weaker), document that `RUN_TSAN=1` requires Catch2 v3.

~3-line change to `scripts/ci_preflight.sh` (+ optional matching tweak in `ci.yml`).

joeytroy:
## Summary
- **#116 — camera family close-order/lock audit**: the ZWO, Player One, and SVBONY cameras carried the same snapshot-then-call class the PR #115 round-3 review found in the switches (id snapshotted under `mutex_`, SDK called after release — a racing disconnect could close the camera in the gap). Every fast register/control call is now a `with_camera` lambda holding `mutex_` across the SDK call (the ToupTek `with_handle` shape); all three disconnects clear state and publish `connected_ = false` **before** the SDK stop/close. Exposure workers and SVBONY's duration-blocking `SVBPulseGuide` keep documented bare snapshots (they must not hold the lock across blocking waits). Every site was mapped before conversion — no site is reached both with and without `mutex_` held, and the `exposure_lifecycle_mutex_ → mutex_ → SDK mutex` order is preserved.
- **ConformU 4.4.0 finding — sticky Error state (ZWO, QHY)**: a live ASI2600MM Pro run hit a one-off `ASI_EXP_FAILED` on the 4×4-bin exposure and the sticky `CameraState::Error` mapping cascaded it into 18 issues. Reproduced the identical 4×4 sequence four times cleanly against the same camera (transient SDK/USB class, not deterministic). ZWO now auto-retries a failed in-flight exposure up to 2× (INDI's ASI policy) then latches — `ImageReady` stays false, `ImageArray` throws, `CameraState` reports Idle since the camera is ready for the next exposure. QHY's identical sticky mapping fixed in the sibling sweep; Player One/SVBONY already resolve to Idle via worker state machines and deadline watchdogs.
- **#117 — TSan zero-test guard**: replaced the Catch2 v3-only `--list-tests | wc -l` count with a run-summary assertion (`test cases: N`, N ≥ 1) in both `scripts/ci_preflight.sh` and the `sanitizers-tsan` CI job — asserts on what actually executed, version-agnostic, `pipefail`-safe.
- **Same-audit bonus**: `get_control_caps_or_throw` returned a reference into the locked caps map (dangling once a reconnect reloads it) — now returns by value (ZWO, SVBONY).

## Changes
- **ZWO / Player One / SVBONY camera drivers** (AlpacaCore): `with_camera` conversion, disconnect reorder, ZWO exposure retry + non-sticky failure latch, QHY state mapping
- **Stress tests**: 6 new `[stress]` cases — ZWO camera and Player One camera storms added to their existing files, new `test_svbony_concurrency_stress.cpp` (23 `[stress]` cases total, zero TSan warnings under the all-vendors TSan build)
- **CI/scripts**: run-summary guard in `ci_preflight.sh` + `ci.yml`
- **AGENTS.md**: `with_camera` siblings recorded next to the `with_handle` reference; blocking-call exemption and the return-by-reference caps trap documented
- **CHANGELOG.md**: three entries under 3.0.0 UNRELEASED

## Test plan
- [x] Local CI pre-flight green: clang-format, unicode, build+test vendors OFF and ON (261 cases / 1896 assertions), clang-tidy, cppcheck, shellcheck, zizmor
- [x] TSan `[stress]` suite: 23 cases, exit 0, zero ThreadSanitizer warnings
- [x] Hardware: ConformU 4.4.0 clean on ZWO ASI2600MM Pro and Player One (this working tree); live 4×4-bin reproduction ruling out a deterministic bin-4 failure
- [x] SVBONY changes are shape-identical to the two hardware-validated drivers and covered by the new TSan stress cases (no SVBONY hardware available)

Closes #116. Closes #117.


joeytroy:
All three round-1 findings verified and fixed in the latest push:

1. **`get_image_array` blocking under `mutex_`** — confirmed; the download now runs on a bare id snapshot outside the lock (the same documented exemption class as the exposure workers), so `stop_exposure`/`abort_exposure` can always acquire `mutex_` and issue the SDK stop mid-transfer. Added a buffer-size re-check after relocking: if ROI/bin changed during the unlocked window, the call throws `InvalidOperation` instead of building an image from a mismatched buffer. Worth noting the pre-PR code had the same hold (id snapshot immediately followed by a lock held across the download) — the finding is real either way, and this closes it properly.

2. **Detached pulse threads capturing `this`** — confirmed; both drivers' turn-off threads now capture only the id snapshot and call the SDK directly, so destruction mid-pulse can't dereference freed members. Player One's second (flag-clear) thread — the same hazard — is deleted outright: `IsPulseGuiding` now self-clears from a `pulse_guiding_end_` timestamp in the getter, adopting the ZWO shape.

3. **Retry not resetting the exposure clock** — confirmed; `poll_exposure_status_locked` now stamps `last_exposure_start_` when it re-triggers, so `PercentCompleted` tracks the retried exposure instead of reporting an immediate false 100%.

Validation on the push: 261 unit tests / 1896 assertions, 23 `[stress]` cases under the all-vendors TSan build with zero warnings, clang-format clean.

joeytroy:
All three round-2 findings verified and fixed in the latest push:

1. **SVBONY detached flag-clear thread capturing `this`** — confirmed, a genuine miss in the round-1 sweep. The thread is deleted outright.
2. **SVBONY reporting guiding after the pulse instead of during** — confirmed; since `SVBPulseGuide` is one blocking call, the end timestamp and flag are now published *before* the call, the flag clears right after it returns (and on throw via catch-rethrow), and the timestamp getter covers readers while the call is in flight. No thread needed at all.
3. **Flag/end-time ordering race** — confirmed for Player One (introduced in round 1) and fixed; the same ordering existed pre-PR in ZWO and was swept in the same commit. All three drivers now write `pulse_guiding_end_` under `mutex_` *before* storing the flag, so a reader that observes the flag can never see a stale end time and self-clear the pulse early.

Validation on the push: 261 unit tests / 1896 assertions, 23 `[stress]` cases under the all-vendors TSan build with zero warnings, clang-format clean.

joeytroy:
Round-3 finding confirmed and fixed in 570a7db — and it was a sharp catch: the round-1 commit message claimed this fix, but the edit automation aborted on an earlier hunk mismatch that round and the recovery pass only reapplied the pulse-thread fixes, so the retry branch never got the timestamp reset. `poll_exposure_status_locked` now stamps `last_exposure_start_` when it re-triggers (member made mutable alongside the other retry state), so `PercentCompleted` tracks the retried exposure. Validated: 261 unit tests / 1896 assertions, clang-format clean.

joeytroy:
Round-4 findings 1–3 verified and fixed in the latest push; finding 4 declined with evidence.

1. **Stale image cached after a racing stop** (CONFIRMED) — fixed: `get_image_array` re-checks `last_exposure_valid_ && image_ready_` under the post-download lock and throws `InvalidOperation` instead of caching a cancelled exposure's pixels.
2. **`get_is_pulse_guiding` unlocked clear racing a new pulse** (CONFIRMED) — fixed in all three drivers: expiry check + flag clear now happen under one `mutex_` hold, and `pulse_guide` publishes the flag inside the same locked block that writes the end timestamp, so a reader can never observe an inconsistent flag/end pair from either side.
3. **Retry burn-down across independent pollers** (PLAUSIBLE, accepted) — fixed with a 250 ms minimum retry interval: within the window the poll reports `Working` and a later poll performs the retry, so back-to-back `CameraState`/`ImageReady` polls can no longer exhaust the budget before the transient clears.
4. **`mutex_` held across `stop_exposure`/status/retry-trigger USB calls** (PLAUSIBLE, declined) — `ZWOSDKWrapper` already serializes **every** SDK call behind its own single mutex (`zwo_sdk_wrapper.cpp`), so a stalled control transfer blocks all concurrent SDK operations at the wrapper regardless of whether the driver lock is held; releasing `mutex_` around these calls would not change the serialization point, it would only reopen the snapshot-then-call windows this PR exists to close. These are single USB control transfers (the AGENTS.md shape-(a) "fast register/control" class), unlike the bulk image download, which is the one call the PR deliberately runs unlocked.

Validation on the push: 261 unit tests / 1896 assertions, 23 `[stress]` cases under the all-vendors TSan build with zero warnings, clang-format clean.

joeytroy:
Both round-5 findings accepted and fixed in the latest push:

1. **Retry omitted the exposure-duration register** — the register value is now stored (`exposure_reg_us_`) at `start_exposure` and re-applied via `set_control_value` immediately before the retry re-trigger, matching the INDI ASI driver's re-set-controls-before-retry policy.
2. **Post-download guards insufficient for the reconnect + short-second-exposure window** — added the suggested sequence token: `exposure_seq_` is bumped by `start_exposure` and by every exposure-state reset (including disconnect), snapshotted under the first lock in `get_image_array`, and validated after the unlocked download. Stale pixels can now never be cached as a fresh image regardless of how the flag/buffer-size checks read.

Validation on the push: 261 unit tests / 1896 assertions, 23 `[stress]` cases under the all-vendors TSan build with zero warnings, clang-format clean.

joeytroy:
Round-6 findings addressed in the latest push — one accepted in part, one accepted in full:

1. **`mutex_` across blocking SDK calls** — accepted for the public `stop_exposure`: the abort intent (retry cancellation, image-state discard, `exposure_seq_` bump) is now recorded under the lock first, then `ASIStopExposure` runs on a bare id snapshot outside it, so a USB hang in the stop can no longer freeze the pure-state readers. **Declined for the disconnect path**: closing under `set_connected`'s lock is the deliberate, documented teardown shape across all 26 drivers (the AsyncConnectable base documents "deferred disconnect holds the lock through hardware teardown" as an accepted cost from the PR #115 review) — and a hung close freezes the driver's SDK access at the wrapper mutex regardless, so releasing the driver lock there changes consistency guarantees without changing the stall. Diverging one driver from the fleet-wide disconnect shape is the kind of asymmetry these audits exist to remove.

2. **Stop before retry** — accepted: the retry path now issues a best-effort `stop_exposure` (try/caught) before re-applying the exposure register and re-triggering. If the SDK tolerates a start from the failed state it's a harmless no-op; if it doesn't, this prevents the instant-fail retry burn-down.

Validation on the push: 261 unit tests / 1896 assertions, 23 `[stress]` cases under the all-vendors TSan build with zero warnings, clang-format clean.

joeytroy:
Both round-7 findings accepted and fixed in the latest push, plus the clarifying note:

1. **Retry SDK calls under `mutex_`** — `poll_exposure_status` is now self-locking: the status read and the retry's three SDK calls run on a bare id snapshot outside the lock; only the retry *decision* (pacing + budget decrement, one winner per window) and the clock restart happen under it. A post-retrigger sequence check stops the new exposure if a stop/abort/disconnect raced the unlocked window, so the round-6 stop-vs-retry guarantee survives the unlock. All three callers reshaped to invoke it without holding `mutex_`.
2. **`ImageReady` stuck true after a throwing download** — readiness is now marked only by a completed download; the pre-download status check no longer sets the flag, and the post-download guard relies on the sequence token (which stops and disconnects bump) instead of the flag.
3. **CI grep note** — added the one-line comments in both `ci.yml` and `ci_preflight.sh` stating the grep is only the zero-test guard and pipefail on the binary exit code is the pass/fail gate.

Validation on the push: 261 unit tests / 1896 assertions, 23 `[stress]` cases under the all-vendors TSan build with zero warnings, clang-tidy clean on the changed lines, shellcheck and clang-format clean.

joeytroy:
All round-8 findings addressed in the latest push. Findings 1 and 3 shared a root cause — the retry's unlocked SDK burst wasn't serialized against user triggers — so rather than patching each symptom, the fix is structural:

1+3. **New `exposure_trigger_mutex_`** (lock order: trigger → `mutex_` → SDK mutex): user `start_exposure`, user `stop_exposure`, and the retry burst hold it across their SDK calls, so a retry can never interleave with or cancel a user's new exposure (finding 1), and two retry bursts can never overlap even when the USB transfers outlast the 250 ms pacing window (finding 3). Status pollers `try_lock` and report `Working` while any trigger is in flight — they never block on USB, preserving the round-7 liveness property. The post-retrigger sequence check stays for the disconnect case, which by design does not take the trigger mutex.
2. **`get_image_ready` seq guard** — the poll's result is only committed if `exposure_seq_` is unchanged and `last_exposure_valid_` still holds after relocking, so a racing stop's `image_ready_ = false` can no longer be overwritten by a stale `Success`.
4. **Pulse state on disconnect** — `reset_exposure_state_locked` now clears `pulse_guiding_`/`pulse_guiding_end_` in all three drivers (swept to Player One and SVBONY, which had the same gap).

Validation on the push: 261 unit tests / 1896 assertions, 23 `[stress]` cases under the all-vendors TSan build with zero warnings, clang-tidy clean on the changed lines, clang-format clean.

joeytroy:
Both round-9 findings fixed in the latest push:

1. **Pulse flag stuck true for fire-and-forget clients** — accepted, with one correction to the suggested fix: storing through the member from the detached thread would reintroduce the `this`-lifetime hazard round 1 removed (the thread can outlive the driver). Instead, `pulse_guiding_` is now a `shared_ptr<atomic<bool>>` the thread **co-owns** — it clears the flag unconditionally after the pulse, and the control block keeps the flag alive even if the driver is destroyed mid-pulse. Fire-and-forget clients get the unconditional clear, pollers keep the timestamp self-clear, and destruction stays safe. Applied to ZWO and Player One; SVBONY's blocking call already clears synchronously.
2. **Unguarded retry re-trigger** — accepted: `set_control_value` + `start_exposure` in the retry burst now carry the same try/catch as the stop above them, logging and surfacing `Failed` to the poller instead of letting a disconnect-window exception escape through a state getter as a raw 500.

Validation on the push: 261 unit tests / 1896 assertions, 23 `[stress]` cases under the all-vendors TSan build with zero warnings, clang-tidy clean on the changed lines, clang-format clean.

joeytroy:
Round-10 finding accepted and fixed in the latest push: `start_exposure`'s register write + trigger now run on a bare id snapshot outside `mutex_`, matching `stop_exposure`'s shape and justification. This is safe now in a way it wasn't when the held-lock shape was written: the round-8 `exposure_trigger_mutex_` (held from function entry) already serializes the pair against stops and retry bursts, so nothing can interleave between the register write and the trigger — the original reason for holding `mutex_` across them. Bookkeeping moves to its own lock hold after the trigger. Validation: 261 unit tests / 1896 assertions, 23 `[stress]` cases under all-vendors TSan with zero warnings, clang-tidy and clang-format clean.

PR #118, opened 2026-07-05
Relicense SSPL-1.0 → AGPL-3.0-or-later (with vendor-SDK linking exception)

joeytroy:
## Summary
- Relicenses the entire repository from SSPL v1 to **AGPL-3.0-or-later**: SSPL is not OSI-approved and is classified non-free by Debian/Fedora, permanently blocking AlpacaBridge from distro archives and third-party apt repos — while AGPL §13 already covers the actual threat model (a closed-hardware vendor embedding the bridge and serving its Alpaca API over the network triggers the source obligation for their modified version).
- Adds a **GPLv3 §7 additional permission** (vendor-SDK linking exception, appended to `LICENSE`) allowing AlpacaBridge to be combined with the proprietary device-vendor SDKs it links and conveyed, without those SDKs becoming AGPL-licensed — pure AGPL would make redistributing the combined binaries self-contradictory.
- Provenance is clean for a unilateral relicense: git history contains only Joey Troy (two emails, same person) and dependabot[bot] (4 workflow version bumps, no copyrightable contribution).
- The change applies prospectively — revisions already published under SSPL remain SSPL at those historical revisions (historical CHANGELOG entries deliberately untouched).

## Changes
- **LICENSE ×3** (root, AlpacaCore, AlpacaHTTP): canonical AGPL-3.0 text + the vendor-SDK linking exception appended
- **210 source files**: SSPL header paragraph → AGPL paragraph; project names, copyright lines, and all three historical wrap variants handled with zero misses; comment-only change (full unit suite unchanged: 255 cases / 1893 assertions)
- **Docs**: READMEs ×3 (root README also gets a reworded intro and an ARA-style License section with the AGPL link), CONTRIBUTING ×2, AGENTS.md, `.cursor` rules ×3, `.claude` commands ×3 — license prose and embedded header examples updated
- **debian/copyright**: `Files: *` → AGPL-3.0-or-later + exception (referencing `/usr/share/common-licenses/AGPL-3`); vendor stanzas corrected and completed for all six shipped SDK libs — ZWO is MIT (was mislabeled Proprietary), Player One permissive-with-notice, QHY (incl. GPL-2 fxload note), ToupTek, SVBONY
- **AlpacaCore/external/README.md**: new vendored-SDK provenance/license table — ZWO and Player One have written grants; QHY/ToupTek/SVBONY are vendored unmodified from official downloads with no written grant (pre-existing condition, now documented, with a removal-on-objection policy)
[...]

PR #121, opened 2026-07-06
Fix #120: guard ZWO exposure-status read; align ZWO copyright name

### LINKED ISSUE #120: ZWO poll_exposure_status: unguarded get_exposure_status read can surface a raw SDK exception on a racing disconnect
Follow-up from PR #119 round-11 review (noted at approval as pre-existing, non-blocking).

`poll_exposure_status()` in `AlpacaCore/src/vendors/zwo/zwo_camera_driver.cpp` calls `ZWOSDKWrapper::instance().get_exposure_status(id)` without a try/catch, while every other SDK call in the same function (the retry stop/re-apply/re-trigger burst) is individually guarded. A disconnect closing the camera between the id snapshot and the status read can surface an SDK exception uncaught through `get_camera_state` / `get_image_ready` — the router's top-level handler prevents a crash, but the client gets a raw error instead of a well-formed Alpaca response.

Pre-existing exposure (the old direct call in `get_camera_state` had it too), not a regression from PR #119 — but the centralized poll is the natural place for the guard.

**Fix:** wrap the status read in the same try/catch pattern as the retry burst, returning `ZWOExposureStatus::Idle` on exception (one-liner plus catch block).

joeytroy:
## Summary
- **#120** (PR #119 round-11 approval note): `poll_exposure_status` read the SDK exposure status on a bare id snapshot with no guard, while every other SDK call in the same function carries one — a disconnect closing the camera between the id snapshot and the read could surface a raw SDK exception through `CameraState`/`ImageReady` instead of a well-formed Alpaca response. The read now returns `Idle` on a throwing status call (logged at DEBUG), matching the retry burst's guards.
- **PR #118 round-5 approval note**: the `debian/copyright` ZWO mount-docs stanza said "ZWO Inc." while the SDK stanza and the SDK's own `license.txt` say "ZWO Company" — aligned to the `license.txt` name. Bundled here per the trivial-fix convention rather than spending a review cycle of its own.

## Test plan
- [x] Local CI pre-flight green: clang-format, unicode, build+test vendors OFF and ON (261 cases / 1896 assertions), clang-tidy, cppcheck (shellcheck/javascript/zizmor skipped — no matching changes)
- [x] TSan `[stress]` suite: 23 cases, zero ThreadSanitizer warnings

Closes #120.


joeytroy:
Round-1 finding fixed in the latest push, using option (b): the poll's id snapshot now returns `Idle` instead of throwing NotConnected, consistent with the guarded status read one level down — `get_camera_state` maps it to Idle (its own not-connected answer), `get_image_ready` to false, `get_image_array` to a well-formed "Image not ready". Went with (b) over (a) deliberately: storing `connected_ = false` before clearing the id would reorder the disconnect sequence for one window in one driver and diverge from the fleet-wide disconnect shape the #116 audit just made uniform. Validation: 261 unit tests / 1896 assertions, 23 `[stress]` cases TSan-clean, clang-tidy and clang-format clean.

PR #128, opened 2026-07-12
Fix #127: enforce HTTP header-size cap on the terminating recv

### LINKED ISSUE #127: HTTP read_request: header-size cap can be exceeded by up to one recv buffer
Follow-up from the PR #124 review (bot-approved, non-blocking).

In `AlpacaHTTP/src/http/server.cpp` `read_request()`, the `kMaxHeaderBytes` (64 KB) check only runs on loop iterations where the `\r\n\r\n` terminator hasn't been found yet. The recv chunk that *contains* the terminator is appended and then breaks the loop before the size check, so the header block can exceed the nominal cap by up to one recv buffer (~8 KB).

Not exploitable — the total is still bounded and the body path enforces `kMaxBodyBytes` separately — just slightly looser than the stated limit. Tighten by checking the size on the terminating iteration too (or check right after every append, before the break).

Severity: low / cleanup.

joeytroy:
## Summary
- Follow-up to the #124 review. `AlpacaHTTP/src/http/server.cpp` `read_request()` only checked `kMaxHeaderBytes` on iterations that had not yet found the `\r\n\r\n` terminator, so the recv chunk containing the terminator was appended and broke the loop *before* the size check — the header block could exceed the 64 KB cap by up to one recv buffer (~8 KB).
- The check now runs on every iteration. When the terminator is found, only the bytes up to it count as headers (the rest is body); before that, the whole accumulated buffer is header-so-far. The cap is now enforced exactly, still returning `431`.

Closes #127.

## Test plan
- [x] `run_all_tests.sh` — vendors OFF + ON, 273 + 278 tests pass
- [x] clang-format + clang-tidy clean on the changed lines
- Bounded-only correctness fix (no exploitable change); existing routing tests exercise `read_request` indirectly via the server path.

### DIFF
--- a/AlpacaHTTP/src/http/server.cpp
+++ b/AlpacaHTTP/src/http/server.cpp
@@ -325,12 +325,18 @@ bool read_request(util::SocketHandle socket_fd, std::string& raw_request) {
         }
         raw_request.append(buffer, static_cast<std::size_t>(bytes_read));
         header_end = raw_request.find("\r\n\r\n");
-        if (header_end != std::string::npos) {
-            break;
-        }
-        if (raw_request.size() > kMaxHeaderBytes) {
+        // Enforce the header-size cap on every iteration, including the one that
+        // finds the terminator — otherwise the terminating chunk could push the
+        // header block up to one recv buffer past the cap unchecked. When the
+        // terminator is found, only the bytes up to it count as headers (the
+        // rest is body); before that, the whole buffer is header so far.
+        const std::size_t header_bytes = (header_end != std::string::npos) ? header_end : raw_request.size();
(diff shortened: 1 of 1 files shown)


joeytroy:
Added `tests/test_server_socket.cpp` — a socket-level regression test for the header-size (431) boundary this PR fixes (first coverage of the `Server` read path, which `test_routing.cpp` bypasses). Verified it fails against the pre-fix `read_request` and passes with the fix. Full suite: 273 + 279 tests pass. This also closes #129. (Note: the PR description edit is currently blocked by a GitHub Projects-classic API deprecation error, so the body still lists only #127 — I'll close #129 manually on merge if needed.)

PR #139, opened 2026-07-16
Fix QHY PulseGuide blocking, Connect() race, and disconnect hang

htly9981:
## Summary
- Fixes 3 real bugs found while validating QHY miniCam8M with ConformU: PulseGuide blocking the HTTP thread, a Connect()/disconnect race that could leave the camera stuck disconnected, and a disconnect hang caused by SDK calls with no timeout.
- Adds 4 hardware-free regression tests against the shared `AsyncConnectable` base (used by all 26 drivers).
- ConformU 4.4.0 on QHY miniCam8M: 0 errors, 0 issues, 0 timing issues.

## Changes
- **QHY Camera Driver** (AlpacaCore): `PulseGuide`'s `ControlQHYCCDGuide` no longer blocks the HTTP thread for the full pulse duration (moved to the Player One camera's detached-thread pattern); cooler-off and temp-control workers bound their joins (2s) and detach instead of blocking disconnect indefinitely on `SetQHYCCDParam`/`ControlQHYCCDTemp`, which have no SDK-side timeout.
- **QHY SDK Wrapper** (AlpacaCore): the camera handle is now reference-counted (`shared_ptr` with a `CloseQHYCCD` deleter) instead of a raw pointer, so a detached worker's in-flight SDK call can never be invalidated by a concurrent `close_camera()`.
- **AsyncConnectable** (AlpacaCore): TRACE logging at the connect/disconnect task's key decision points; useful for diagnosing this bug class across all 26 drivers that share it.
- **Router** (AlpacaHTTP): the disconnect-completion wait now polls `get_connecting()` alone instead of `get_connected()` -- some drivers correctly clear the connected flag before teardown finishes, which was letting the router report "disconnected" early and causing the client's next `Connect()` to race the still-running task and get silently dropped.
- **AsyncConnectable Unit Tests** (AlpacaCore): 4 new Catch2 test cases against the shared base directly, no hardware needed.
- **Documentation**: AGENTS.md, CHANGELOG.md (UNRELEASED 3.0.1), SUPPORTED-DRIVERS.md updated.

## Test plan
- [x] Local CI pre-flight green: clang-format, clang-tidy, cppcheck, unicode scan, build+test (vendors OFF + ON)
- [x] Unit tests pass (1927 assertions, 277 test cases)
- [x] ConformU 4.4.0 passes on QHY miniCam8M, Linux arm64

## ConformU results
- **arm64**: `AlpacaCore/conformu/QHY/miniCam8M/Linux-arm64.txt` -- 0 errors, 0 issues, 0 timing issues


joeytroy:
@htly9981, please make sure you are using Claude to submit the PR. There are AI skills set up in the repo, and if they are not followed, you can not submit a PR. This allows for continued integration by following the same processes again and again. The agent will take care of what needs to be documented, along with running the ConformU tests and building proper tests for the AlpacaBridge with the hardware for the MiniCam. Documentation is located here 

https://github.com/open-astro/AlpacaBridge/blob/main/docs/development.md#claude-code-skills

joeytroy:
@htly9981, it also looks like you didn't pull down the last changes I made before the commit.

htly9981:
sorry, had some problems, I had to build the filterwheel part. and sorry for lack of basic coding skill and rules. l'm worried about this makes everyone in trouble.

joeytroy:
@htly9981, not to worry; as long as you use Claude and follow the skills, it can build the driver properly for you. You can also copy the SDK from QHY into the repo as well, and it will build the filter wheel driver. Also, for what it is worth, I am still waiting for QHY to send me a bunch of hardware. Once it arrives, I will get to work on the drivers as well.

joeytroy:
@htly9981 I fixed the merge issue; you were just behind on the pull. It's going through the tests now. I will work through the rest of the PR to see it to the end. Are you going to add the filter wheel as well for the Mini Cam8?

joeytroy:
@htly9981 Thanks for this PR — hardware-validated fixes with a clean ConformU 4.4 run and regression tests against the shared base. Overall high-quality work that follows the project's concurrency conventions closely, with excellent comments explaining *why*. The router fix and the shared_ptr-handle design are exactly right. Review below — one crash-level bug and one lifecycle bug on the new detach path need addressing before merge.

*(Note: I merged `main` into this branch to resolve conflicts with the new `pending_connect_` semantics in `AsyncConnectable` — the "connect racing an in-flight disconnect" test now asserts the queued behavior.)*

## Issues

### 1. HIGH — PulseGuide's detached thread can call `std::terminate` (`qhy_camera_driver.cpp`, `pulse_guide`)

```cpp
std::thread([cam_id, qhy_dir, dur_ms, flag = pulse_guiding_]() {
    QHYSDKWrapper::instance().guide(cam_id, qhy_dir, dur_ms);
    flag->store(false);
}).detach();
```

`guide()` throws: `get_handle()` throws `AlpacaException(NotConnected)` if the camera was closed between the handler's connected check and the thread running, and `check_result` throws on any SDK failure. An exception escaping a `std::thread` entry point calls `std::terminate` — a disconnect racing a pulse **crashes the whole server**. The old code only slept, so this is newly introduced. This PR itself fixed exactly this class of problem in `AsyncConnectable::run_connection_task` (catch-all + log). Needs:

```cpp
try { QHYSDKWrapper::instance().guide(cam_id, qhy_dir, dur_ms); }
catch (const std::exception& e) { ALPACA_LOG_WARN("QHY", ...); }
catch (...) { ALPACA_LOG_WARN("QHY", ...); }
flag->store(false);  // must also clear on the throw path
```

The flag must clear on all paths or `IsPulseGuiding` sticks true forever.

### 2. MEDIUM — shared stop/running flags are reused across thread generations, breaking the detach safety story (`qhy_camera_driver.cpp`)

`temp_thread_stop_` / `temp_thread_running_` are member `shared_ptr`s that are `store()`-reset on each spawn, not re-allocated. Sequence:

1. Disconnect: `join_temp_thread` times out, **detaches** the old thread while it's stuck in `control_temp()`.
2. Client reconnects: `start_temp_control_thread` does `temp_thread_stop_->store(false)` — on the **same object the detached thread captured**.
3. The detached thread returns from `control_temp()`, rechecks `stop_flag` → now false → **keeps looping**: two temp threads driving the cooler, and the zombie takes `mutex_` (i.e., touches `this`) every iteration — the exact use-after-free the recheck comment says it prevents, re-armed by the flag reset.

Also `temp_thread_running_`: the new thread stores true, then the zombie's `RunningGuard` destructor stores false — `join_temp_thread` can then `join()` a still-running thread, resurrecting the unbounded hang.

Fix: allocate **fresh** `make_shared` flags per spawn, so each thread generation owns its own stop/running pair and detached zombies keep their old ones (which stay stopped forever). `cooler_off_running_` has the same reuse pattern and should get the same treatment. This only bites on the timeout→detach→reconnect path — rare, but it's precisely the path this PR exists to make safe.

### 3. LOW — `cooler_off_running_` set before thread construction (`set_cooler_on`)

`cooler_off_running_->store(true)` precedes `std::thread(...)`; if the ctor throws (thread limit), the flag is stuck true → every later cooler-off returns early ("worker in flight") and `join_cooler_off_thread` burns its full 2s timeout then hits a non-joinable thread. The `AsyncConnectable` spawn in this same PR handles its rollback correctly — mirror it.

## What's good

- **Router fix** (`router.cpp`): polling `get_connecting()` alone is the correct completion signal, well-argued, and backed by a regression test against the shared base — this protects all 26 drivers, not just QHY.
- **Reference-counted handle** (`qhy_sdk_wrapper.cpp`): clean design; `close_camera`/`open_camera` erase-not-close is right, and every blocking call correctly keeps the `shared_ptr` (not a raw copy) alive across the unlocked region.
- **Tests**: the `TestConnectable` double faithfully mirrors the clear-connected-first driver pattern; the "get_connected is not a completion signal" case pins the router bug precisely.
- **Docs**: AGENTS.md/CHANGELOG/SUPPORTED-DRIVERS updates match repo convention, with the generalizable lessons captured.

## Other notes

- `guide()` holds the wrapper's global mutex for the full pulse (up to 65s in theory — `dur_ms` is uint16), serializing *all* SDK calls behind it (exposure readout, temp control). The comment acknowledges this as pre-existing; fine for now, but worth a follow-up issue since a 65s pulse is representable.
- The timing-instrumentation logs around `SetQHYCCDParam(MANULPWM)` use INFO level; DEBUG would match the surrounding code's verbosity, but harmless.

**Verdict: ⚠️ Issues found** — #1 is a must-fix before merge (server crash on a disconnect-vs-pulse race); #2 should be fixed in the same pass since it undermines the PR's own safety invariant; #3 is a nice-to-have.

htly9981:
Thank you for checking. l'll try to build wheel part of qhyminiCAm8m, maybe next Monday. and fix this. by the way, https://localhost:6800 webui has no qhy selection on filterwheel. Is it web issue? or if l build entire driver include wheel part, filterwheel - qhy will appear???

joeytroy:
@htly9981 if you use the skills you can easily build the driver it's got a step by step walk through with Claude using /driver-build as long as you download the QHY SDK you can put out it in core/external folder you will see the way I setup the rest of the SDK folders. I also completed a new pr, and merged it, so you need to update your repo again with a pull request. I also updated the skills for /submit-pr as well. Once you fix the issues I'll run through it again and make sure it looks good.

And yes the QHY filter wheel will show up.

htly9981:
@joeytroy Conflict with upstream/main resolved and re-pushed (merge commit 1b54a9f) — CHANGELOG.md was the only conflict (version-numbering collision with the released 3.0.2 entry, retitled this branch's unreleased section to 3.0.3). AlpacaCore builds clean and all 1927 assertions / 277 test cases pass locally. CI is showing `action_required` — could you approve the workflow run and kick off a review when you have a chance?

and after that, qhy minicam filer part driver build will be started

joeytroy:
@htly9981 Thanks a bunch for these fixes; everything looks good. I will merge it.

htly9981:
thank you ! you encourage me, my tonight's work will be building filter wheel part driver

PR #142, opened 2026-07-21
Add QHY integrated CFW filter wheel driver with HTTP/UI integration

htly9981:
## Summary
- Adds an AlpacaCore FilterWheel driver for QHY's integrated CFW (e.g. miniCam8M) — a wheel controlled through the SAME physical USB handle as its paired camera, not a separately enumerable device
- Full AlpacaHTTP router/web UI integration, 9 unit test cases, and a clean ConformU 4.4.0 validation on real hardware (Linux arm64, USB)

## Changes
- **QHY Filter Wheel Driver** (AlpacaCore): `QHYSDKWrapper::open_camera()`/`close_camera()` are now reference-counted per camera_id (mirroring ToupTek's camera+thermal-switch sharing), so the camera driver and the new filter wheel driver can connect/disconnect independently while sharing one `OpenQHYCCD`. New wrapper calls `move_cfw()`/`get_cfw_position()` wrap `SendOrder2QHYCCDCFW`/`GetQHYCCDCFWStatus` (ASCII-digit protocol, no `InitQHYCCD` needed); slot count is read from `CONTROL_CFWSLOTSNUM` at connect; `Position` reports −1 while moving per the ASCOM IFilterWheelV3 contract. Both SDK calls release `QHYSDKWrapper`'s global mutex before the blocking SDK round trip (matching `set_param()`/`control_temp()`/`get_single_frame()`), since `GetQHYCCDCFWStatus` is a genuine ~100-130ms hardware call and would otherwise stall every other open QHY camera's SDK calls on every `Position` poll during a filter change.
- **QHY Filter Wheel Support** (AlpacaHTTP): router registration (`qhy` + `filterwheel`), web UI CFW config section (camera index/id, slot-count picker, filter-name editor), config persistence fix (the `qhy` sanitizer branch didn't allowlist `filterNames`), and a routing round-trip test.
- **QHY Filter Wheel Unit Tests**: 9 test cases, 42 assertions covering defaults, metadata, connection-state contracts, and ASCOM value-range/state-machine/unsupported-method behavior.
- **ConformU Validation**: miniCam8M CFW, 4.4.0, Linux arm64, USB — 0 errors, 0 issues, 0 timing issues. A hardware-round-trip timing violation found during validation is fixed with a settled-position cache warmed at `Connect` and invalidated on every `Position` write, falling back to live reads until a read confirms the commanded target.
- **Documentation**: AGENTS.md (lessons learned on the shared-handle CFW pattern, ASCII protocol, and the no-moving-sentinel caching bug), CHANGELOG.md (`3.1.0` — new driver, minor bump), SUPPORTED-DRIVERS.md, docs/architecture.md.

## Test plan
- [x] Local CI pre-flight green: clang-format, unicode scan, `run_all_tests.sh` (vendors OFF + ON), clang-tidy, cppcheck
- [x] Unit tests pass (286/286 AlpacaCore, 292/292 AlpacaHTTP)
- [x] ConformU 4.4.0 passes on Linux arm64 (miniCam8M CFW, real hardware)

## ConformU results
- **arm64**: `AlpacaCore/conformu/QHY/miniCam8M CFW/Linux-arm64.txt`


htly9981:

joeytroy:
@htly9981 I built the code to check it, and I noticed it has a 6 slots filter wheel but not a 5 slots. Also, I assume users know their minicam is 8 slots, so I don't want to specify the miniCam8M CFW, as QHY makes other 8-slot filter wheels and don't want to confuse users.

image

## Review: QHY integrated CFW filter wheel driver

### Overview

Adds an AlpacaCore FilterWheel driver for QHY cameras with an integrated CFW (miniCam8M), where the wheel is controlled through the same USB handle as the camera. To support this, `QHYSDKWrapper`'s handle map becomes reference-counted per camera ID, and two new wrapper calls (`move_cfw`/`get_cfw_position`) wrap the SDK's ASCII-digit CFW protocol. Includes router/web-UI integration, a config-persistence fix (`filterNames` allowlisting), 9 unit tests, a routing round-trip test, a clean ConformU 4.4.0 hardware run, and thorough doc updates.

### Overall assessment

This is a well-executed PR. The hard-won hardware findings (no moving sentinel in `GetQHYCCDCFWStatus`, the ~100–130 ms round trip, the ASCII-digit protocol) are documented at the point of use, in AGENTS.md, and in SUPPORTED-DRIVERS.md consistently. The settled-position cache design — cache only when no move is pending, re-cache only once a live read equals the commanded target, invalidate before issuing the move — is correct against the observed hardware behavior, and the earlier failure mode it fixes is explained honestly. Locking follows the repo's established patterns (fetch the `shared_ptr` under the global mutex, call the SDK unlocked; drop-the-reference close via the `shared_ptr` deleter). Error paths in `set_connected` are careful: open rolled back on connect failure, driver state cleared before close on disconnect. Test evidence is strong (ConformU 0/0/0 on real hardware plus unit and routing tests).

### Findings

#### 1. Stale-handle recovery regression in `open_camera` (medium — worth a deliberate decision)

Previously `open_camera` unconditionally erased any existing entry and re-opened, so a reconnect after an unplug/replug always got a fresh `OpenQHYCCD`. Now (`qhy_sdk_wrapper.cpp`), if any owner still holds the entry (`open_count > 0`), a new opener just bumps the count and reuses the existing handle.

**Scenario:** camera driver and CFW driver both connected → USB device drops and re-enumerates → user disconnects/reconnects only the camera. The reconnect silently reuses the dead handle because the still-"connected" CFW driver pins `open_count ≥ 1`, and every subsequent SDK call fails until *both* owners disconnect. The old self-healing behavior is gone whenever a second owner exists.

If ToupTek's `open_shared_by_id` has the same property this may be an accepted trade-off, but it deserves a comment (or a liveness check on reuse) acknowledging it.

#### 2. `move_cfw` breaks for positions ≥ 10 (low)

`char order = '0' + position` is only valid for slots 0–9, and the web UI allows a custom slot count up to 16 while `set_position` accepts anything below the hardware-reported `slot_count_`. If a QHY wheel ever reports >10 slots, position 10 sends `':'`. QHY's lineup maxes at 9 slots today, so this is theoretical — but a cheap guard in `move_cfw` (`position > 9` → `InvalidValue`) would make the constraint explicit rather than silent.

#### 3. `get_position` holds the driver's `mutex_` across the ~130 ms SDK call (informational)

The comment justifies this (prevents a concurrent disconnect invalidating `camera_id_` mid-call, matches the ZWO/ToupTek pattern), and it's fine for correctness — but note the side effect: `get_name`, `get_unique_id`, `get_names`, `get_focus_offsets`, and `Connected = false` all stall up to ~130 ms behind an in-flight Position poll. Acceptable for a filter wheel; just confirming it's intentional.

#### 4. `focusOffsets` are not persisted (low)

The sanitizer fix allowlists `filterNames`, but offsets set via the Alpaca API are lost on save/restart. ConformU shows all-zero offsets so nothing breaks today; flagging for consistency with the `filterNames` fix — if other vendors' wheels behave the same, ignore.

#### 5. Cosmetic

The clang-format artifact in `qhy_filterwheel_driver.cpp` (the `cached_position_` / `.reset();  // comment` line split in the connect path) reads awkwardly — moving the comment above the statement would let it format normally.

### Code quality, conventions, tests, security

- **Style** matches the codebase closely: same license header, locking idioms, `AlpacaException` error-code discipline, ASCOM validation-precedence rules (negative position → `InvalidValue` even when disconnected, upper bound → connected-only), and the ToupTek shared-handle precedent it cites.
- **Test coverage** is appropriate for a hardware driver: unit tests cover everything testable without hardware (state machine, error codes, DeviceState contract, metadata), the routing test guards the exact persistence bug that was fixed, and ConformU covers the hardware paths. The connected-side cache logic (`pending_target_` state machine) is only exercised on hardware — a mockable SDK seam would let it be unit-tested, but that's a larger refactor than this PR should carry.
- **Security:** no concerns — input is range-validated before reaching the SDK, `filterNames` is type-checked in the router, and the fixed-size `status[64]`/`id_buf[64]` buffers are used safely.

### Verdict

If you can resolve the findings.

htly9981:
@joeytroy thanks for the review — pushed 06cbbee addressing this:

- **Slot picker**: fixed the `6/7/8/9` → `5/7/8/9` typo (QHY has no 6-slot CFW), and dropped the `(miniCam8M CFW)` qualifier from the 8-slot option so it reads as a plain "8 slots" like the other vendor pickers, no longer implying it's the only 8-slot QHY wheel.
- **Finding #1 (stale-handle reuse)**: confirmed ToupTek's `open_shared_by_id` has the identical property, so documented it explicitly as an accepted trade-off in `open_camera()` (no liveness check on reuse yet — noted where one would go if it ever bites).
- **Finding #2 (`move_cfw` positions ≥ 10)**: added an explicit `position > 9` → `InvalidValue` guard in `move_cfw()` rather than leaving the ASCII-digit overflow silent.
- **Finding #5 (cosmetic)**: moved the comment above the `cached_position_.reset()` statement so it formats normally.
- **Finding #3** (mutex held across the ~130ms poll) and **#4** (`focusOffsets` not persisted): left as-is per your notes — #3 is intentional/matches the ZWO/ToupTek pattern, and #4 is consistent with every other vendor's filter wheel (none persist `focusOffsets` today).

AlpacaCore (88 assertions) and AlpacaHTTP routing tests both green after the changes.

joeytroy:
@htly9981  I have fixed the repo to support the bot for users when they submit PR's now. This PR looks good. Can you please rerun ConformU for the camera and also the filter wheel again and push it back to the repo, as I have updated the SDK from 25.09.29 to 26.06.04. I just want to make sure it's still all working as I still do not have the hardware from QHY yet.

htly9981:
ok, i will, but I have to work overtime .grrrrr.....give me some more time to do it

joeytroy:
@htly9981 I implemented the Wanderer Astro Rotator so I had to fix some things with your PR. Looks good now. Just run the ConformU for the camera and filter wheel so we know it's good and push those reports up. As long as it's approved status I will merge tomorrow.

htly9981:
@joeytroy ConformU re-validation is pushed (87130ae) — camera and CFW both pass 4.4.0 against the new SDK (0 errors, 0 issues, 0 timing issues on real miniCam8M hardware), plus the stale `25.09.29.11` note in `SUPPORTED-DRIVERS.md`'s FilterWheel section is fixed and CHANGELOG now credits your connect-failure rollback fix.

All CI is green **except** the `review` check, and it's not a driver issue — it's the fork-review actor gate:

- `claude-review.yml` is correctly label-gated (`safe-to-review` is on this PR) and grants `pull-requests: write` to the token.
- But `claude-code-action` itself does an independent check of whether **`github.actor`** (the person whose push triggered the run) has write access to `open-astro/AlpacaBridge`. Your own pushes to this PR (e.g. `7d42072`) pass that check since you have write access; my push just now (`87130ae`, actor `htly9981`) only has read access upstream, so the action refuses with "Actor does not have write permissions to the repository" before it can post a review, and the "Assert the review comment posted" step then fails the job.
- Every other check (build+test, sanitizers, CodeQL, clang-tidy, cppcheck, clang-format, etc.) is passing on this commit.

Could you re-run the failed `review` job from your side? If re-running attributes `github.actor` to whoever clicks re-run, that alone should clear it. If not, the actor-permission check inside `claude-code-action` probably needs to also accept the `safe-to-review` label as sufficient authorization, same as the job-level `if` already does.

joeytroy:
@htly9981 can you log onto the discord server. A user is having an issue using NINA with the minicam8 and since I do not have one I cannot test it. I would very much appreciate it.

PR #143, opened 2026-07-21
Add Gemini Astro Flat Panel Cover Lite driver

htly9981:
## Summary
- Adds a CoverCalibrator driver for the Gemini Astro Flat Panel Cover Lite, a light-only flat panel with no motorized cover.
- Protocol was reverse-engineered from the vendor's Windows control app (no SDK or published spec) — ASCII commands over USB serial.
- ConformU 4.4.0 validated on Linux arm64: 0 errors, 0 issues, 0 timing issues.

## Changes
- **Gemini Flat Panel Driver** (AlpacaCore): `gemini_flatpanel_protocol_wrapper` implements the `>X#`/`>Xnnn#` command set at 9600 baud 8N1 (replies `*` + echoed letter + payload + `#`, with `>S#` as the one exception returning three single-digit flags). `gemini_flatpanel_driver` wraps it as a `CoverCalibratorDriver`. Auto-detection scans `/dev/serial/by-id/` for CH340/CH341/generic USB-serial names plus `Espressif` (the panel's controller is an ESP32-class board on the native USB-serial/JTAG stack, not a CH340/CH341 adapter like the Gemini focuser), probes with the `>H#` identity handshake, and falls back to `/dev/ttyUSB0`–`9`. No motorized cover: `OpenCover`/`CloseCover`/`HaltCover` throw `MethodNotImplemented`, `CoverState` always reports `NotPresent`.
- **Gemini Flat Panel Device Support** (AlpacaHTTP): router registration for the `covercalibrator` device type under the `gemini` vendor (auto-detect or manual serial config), web UI vendor/device-type gating and config fields — Gemini now covers two device types (Focuser + CoverCalibrator) from one vendor config block, following the same show/hide pattern already used for iOptron's mount/switch split.
- **Gemini Flat Panel Unit Tests**: 9 test cases, 48 assertions covering defaults, metadata, disconnected behavior, unsupported actions, no-cover reporting, value-range validation, and the connect/disconnect state machine.
- **ConformU Validation**: `AlpacaCore/conformu/Gemini/Astro Flat Panel Cover Lite/Linux-arm64.txt` — 4.4.0, 0 errors/issues/timing issues.
- **Documentation**: CHANGELOG.md (`3.1.0` — minor bump for the new driver), SUPPORTED-DRIVERS.md (new CoverCalibrator table entry + driver notes), docs/architecture.md (Gemini's device-types cell now lists both Focuser and CoverCalibrator).

## Test plan
- [x] Local CI pre-flight green: `run_all_tests.sh` (vendors OFF + ON), clang-format, unicode scan, clang-tidy, cppcheck
- [x] Unit tests pass (`ctest` — 292/292, incl. the 9 new flat panel cases)
- [x] ConformU 4.4.0 passes on Linux arm64 (0 errors, 0 issues, 0 timing issues)
- [ ] Web UI configuration exercised in a browser (not covered by this pass — worth a manual smoke test against real hardware)

## ConformU results
- **arm64**: `AlpacaCore/conformu/Gemini/Astro Flat Panel Cover Lite/Linux-arm64.txt`


htly9981:

joeytroy:
## Review: Gemini Astro Flat Panel Cover Lite driver

### Overview

Adds a CoverCalibrator driver for the Gemini Astro Flat Panel Cover Lite, a light-only flat panel (no motorized cover) with a reverse-engineered ASCII serial protocol (`>X#` / `>Xnnn#` at 9600 8N1). The stack follows the established vendor pattern: a pimpl protocol wrapper owning the serial port, a driver class on top of `AsyncConnectable`, auto-detection via `/dev/serial/by-id/` probing, router registration sharing the `gemini` vendor with the existing focuser, a web UI sub-section split (iOptron pattern), 9 unit tests, and a clean ConformU 4.4.0 hardware run.

### Overall assessment

Strong PR, consistent with the repo's conventions. The reverse-engineering findings are unusually well documented — the `>S#` three-flag exception (and why the generic trailing-int parser would mis-read `*S011#` as "on"), the 22-char `>H#` reply/buffer-size trap, and the ESP32-native-USB vs CH340 distinction are all recorded at the point of use and in AGENTS.md. ASCOM contract details are handled correctly: `CalibratorOn(0)` stays Ready, range validation precedes the connection check, cover methods throw `MethodNotImplemented` unconditionally with `CoverState = NotPresent`, and the commanded-brightness cache with rollback-on-throw mirrors the WandererCover lesson. The state seed from a live read at connect (so reconnecting to an already-lit panel reports the truth) is a nice touch.

### Findings

#### 1. Auto-detect can misidentify the Gemini focuser as a flat panel (main finding — requested change)

`probe_port()` accepts *any* well-formed `#`-terminated reply to `>H#`, and the candidate patterns deliberately include CH340/CH341/`USB_Serial`/`1a86` — which is exactly what the Gemini **focuser** enumerates as (MyFocuserPro2 replies are also `#`-terminated). With both devices plugged in, the flat the focuser's port if the focuser's MCU echoesanything ending in `#`, and `panelIndex` ordering becomes dependent on directory-iteration order over mixed devices.                    
The handshake reply is known exactly (`*HGeminiFlatPanelLite#`), so the fix is cheap: **require the `*H` prefix (or the full string) in `probe_port()`** instead of "any `#`-terminated reply"s looser check is defensible for a manually configuredport, but the *discovery* path needs to discriminate. AGENTS.md's own new warning ("don't assume shared protocol or auto-detect matching") argues for this.

#### 2. Hidden web-UI fields aren't disabled — inconsi

The focuser and flat panel sub-sections share the `gemcuser section keeps `name="connectionType"` /`name="portPath"` / `name="baudRate"` inputs while merely hidden (`display: none`), whereas PR #142's equivalent split calls   `setFieldGroupEnabled()` to disable the inactive grouply overwrites `connectionType` for the flat panel branch so nothing breaks today, but hidden focuser inputs could leak stale `portPath`/`baudRate` into a flat panel config if the form serialization ever changes. Using `setFieldGroupEnableelds()` would match #142 and remove the hazard.

#### 3. Stale/contradictory doc text (minor)

- The web UI help text says "the exact response formatt real hardware" — but the PR elsewhere states theformat is hardware-confirmed and ConformU-validated. Drop or soften it.
- `gemini_flatpanel_driver.h`'s doc for `create_geminians CH340/CH341 USB-serial adapters" — omits theEspressif pattern that is the actual tested path.

#### 4. Small code nits (take or leave)

- `set_brightness()` sizes its command buffer with `kMaxResponseLen` — works, but the constant's name doesn't describe the use; a `kMaxCommandLen` would read better.
- In `probe_port()`, the success path re-runs `tcgetattr`/clear-HUPCL/`tcsetattr` although HUPCL was already cleared during setup — harmless but redundant.
- Auto-detect worst case is slow: per candidate port, up to 2 attempts × (2 s/1 s boot wait + up to 5 s read timeout). Fine on the
background connect thread, but with several non-panel ect could take tens of seconds.

#### 5. Release logistics

Both this PR and #142 rewrite the same `## [3.0.3] - UNGELOG section (and both touch`AlpacaCore/tests/CMakeLists.txt`) — whichever merges second will conflict; plan the rebase.

### Code quality, conventions, tests, security

- **Conventions:** matches the codebase — license headers, `AsyncConnectable` destructor contract (shutdown before member teardown, qualified `set_connected` call), pimpl wrapper with a  shared with the focuser, ASCOM error-code disciplinethroughout.
- **Tests:** the 9 unit cases cover everything testabl the subtle contracts (MethodNotImplemented whiledisconnected for cover methods, InvalidValue-before-NotConnected precedence, DeviceState omitting throwing entries). A router-level
round-trip test (like #142's) is absent — the `panelIntested; a small routing test would guard it.
- **Performance:** serial I/O under `state_mutex_` is ~100 ms per command, consistent with other serial drivers; ConformU timing all green.
- **Security:** no concerns — bounded reads (`kMaxRespo-long throw), validated brightness range before hitting the wire, no untrusted string formatting.

### Verdict                                                                                                                                
Please resolve the findings

htly9981:
@joeytroy thanks for the review — pushed 17522ca and 49cb65e addressing this:

- **Finding #1 (main, requested change)**: `probe_port()` now requires the known `*H` reply prefix instead of accepting any well-formed `#`-terminated reply. Pulled the check out into a standalone `is_flatpanel_handshake_reply()` (declared in `gemini_flatpanel_protocol_wrapper.h`) so the string-matching logic is unit-tested (7 sections: the confirmed real reply, this driver's own `>V#`/`>S#`-style replies, a plausible non-flat-panel `#`-terminated reply standing in for the focuser's MyFocuserPro2 firmware, and short/empty/malformed input) — `probe_port()` itself still needs real hardware to verify end-to-end, since it opens an actual serial port. I don't have a focuser + panel plugged in together to confirm the live discrimination, so that scenario is still only reasoned about, not hardware-tested.
- **Finding #2**: the focuser/flat-panel sub-sections now call `setFieldGroupEnabled()` on the hidden group, matching #142's QHY camera/filterwheel split, so stale `portPath`/`baudRate`/`connectionType` can't leak across device types.
- **Finding #3**: dropped the stale "may still need tuning against real hardware" web UI help text (it's ConformU-validated), and `create_gemini_flatpanel_by_index()`'s doc comment now mentions the Espressif pattern instead of only CH340/CH341.
- **Finding #4**: renamed the command-buffer constant to `kMaxCommandLen`, dropped the redundant tcgetattr/tcsetattr re-run in `probe_port()`'s success path (HUPCL was already cleared during setup). Left the auto-detect worst-case-timing note as-is since you called it acceptable.
- **Tests**: added the routing round-trip test you flagged as missing, guarding `panelIndex` persistence.
- **Finding #5 (release logistics)**: noted — will rebase CHANGELOG.md/AlpacaCore/tests/CMakeLists.txt against whichever of #142/#143 merges first.

AGENTS.md records the auto-detect discrimination lesson. Full AlpacaCore suite (1984 assertions) and AlpacaHTTP routing tests green after the changes.

PR #159, opened 2026-07-28
Fix QHY CFW mid-move position spam and stale web-UI config edits

htly9981:
## Summary
- Fixes a QHY integrated CFW (color filter wheel) bug where `Position` polled mid-move could return a real but unrelated transit slot instead of a clean "moving" signal — NINA (and any other Alpaca client) polling during a filter change could misread an intermediate slot as an erroneous/mismatched arrival before the wheel had actually settled.
- Fixes a web-UI bug where editing a device could silently resubmit and persist a stale cached configuration, discovered while diagnosing the above (a browser tab left open across an out-of-band config change would still show old data, and Edit → Save would clobber the newer config).

## Changes
- **QHY Filter Wheel Driver** (AlpacaCore, `qhy_filterwheel_driver.cpp`): `GetQHYCCDCFWStatus` has no distinct "moving" sentinel — it reports the wheel's actual passing slot throughout the physical rotation (e.g. a commanded 4→3 move observed reporting `4,5,6,-1,0,1,2,3` as it took the short way round). `get_position()` previously passed every live mid-move reading straight through; a live read taken while a move is pending is now masked to `-1` unless it equals the commanded target, matching ToupTek AFW's `Position` semantics. Verified on real miniCam8M hardware: 20 rapid polls during a 5→1 move returned `-1` throughout, then settled cleanly on `1`.
- **Web UI** (AlpacaHTTP `app.js`/`index.html`): `handleEditDeviceClick` now re-fetches the device list (`loadDevices()`) before opening the Edit form, instead of trusting the `currentDevices` cache that's only refreshed on page load or after a same-tab submit. Cache-bust bumped to `v=15`.
- **Documentation**: `AGENTS.md` and `SUPPORTED-DRIVERS.md` QHY CFW notes updated to describe the masking behavior. `CHANGELOG.md`: added a new `[3.1.1] - UNRELEASED` section (patch bump — bug fix only) with two `### Fixed` entries, and collapsed the now-released `[3.1.0] - 2026-07-27` section into the `<details>` form.
- Branch also merges in `upstream/main` (3.1.0 had been released while this branch was based on an older commit) to keep history current.

## Test plan
- [x] Local CI pre-flight green: `scripts/ci_preflight.sh` — clang-format, unicode scan, build+test vendors OFF/ON (336 tests, 100% pass), clang-tidy, cppcheck all PASS. JavaScript syntax check SKIPPED locally (no `node` available in this sandbox); the small diff (added `async`/`await`) was manually reviewed and is valid syntax — GitHub Actions CI has `node` and will run this gate for real.
- [x] Existing QHY filter wheel unit tests pass unchanged (9 test cases, 42 assertions)
- [x] Manually verified on real miniCam8M hardware: rebuilt and redeployed to a live AlpacaBridge instance, reconfigured filter names, cycled through all 8 filter positions, and rapid-polled `Position` during a multi-slot move — confirmed `-1` throughout transit and a clean settle at the commanded target, no stray intermediate slot values leaked.
- [ ] No new ConformU run was performed for this fix (behavioral change is only observable during a live in-flight move, which ConformU's own timing/position tests already exercise as part of the existing clean `miniCam8M CFW` report)


htly9981:

htly9981:
Addressed the review feedback on `handleEditDeviceClick` in `f8c2f1c`.

**Issue**: `index` was captured from `button.dataset.deviceIndex` before the `await loadDevices()` reload, then used to index into the freshly re-sorted `currentDevices` afterward. An out-of-band change that altered the device list's composition or sort order could make the index point at a different device post-reload — clicking Edit on device A could open (and on Save, overwrite) device B.

**Fix**: capture the clicked device's `DeviceType`/`DeviceNumber` before the reload, then look the device up by that stable identity in the refreshed list instead of by position.

Also bumped the `app.js` cache-bust to `v=16` and updated the `CHANGELOG.md` entry to describe both the original stale-cache fix and this follow-up fix.

Not changed: `handleDeleteDeviceClick` has the same positional-index pattern, but per the review it's pre-existing and not reload-triggered, so left as-is.

PR #169, opened 2026-08-03
Add Gemini Astro Automatic FlatPanel v2 driver (ConformU 4.4.0)

Knetus56:
## Summary
- New CoverCalibrator driver for the **Gemini Astro Automatic FlatPanel v2** — a second, distinct Gemini flat panel model with a real motorized cover, alongside the existing Astro Flat Panel Cover Lite (light-only).
- No vendor docs, protocol spec, or hardware were available up front; the driver was implemented from INDI's open-source `GeminiFlatpanelRev2Adapter` (`indilib/indi`) and then validated against a real unit.
- ConformU 4.4.0 validated on Linux arm64: **0 errors, 0 issues, 0 timing issues**.
- Fixes a real timing bug ConformU caught: `CalibratorOn`/`CalibratorOff` could block for 6+ seconds behind an in-flight cover move, since both share one physical-serial-port mutex in the protocol wrapper.

## Changes
- **Gemini FlatPanel v2 Driver** (AlpacaCore): `GeminiFlatPanelV2Driver` alongside the existing Lite driver in the same vendor files, selected via a `flatPanelModel` config field (`"lite"`/`"v2"`) on the same `gemini`+`covercalibrator` router slot. `OpenCover`/`CloseCover` block the wire for up to 30s waiting for the hardware's exact completion reply (`*OOpened#`/`*CClosed#`) and run on a background thread so the ASCOM async initiator returns immediately. `HaltCover` has no hardware equivalent on this revision — it only stops the driver from *reporting* movement.
- **Gemini FlatPanel v2 Support** (AlpacaHTTP): router registration, web UI model selector (`gemini-flatpanel-model`) toggling between Lite/v2 config sub-sections, shared `panelIndex` auto-detect index namespace with the Lite model.
- **Gemini FlatPanel v2 Unit Tests**: 9 test cases, 52 assertions covering defaults, metadata, connection-state contracts, and ASCOM value-range/state-machine/unsupported-method behavior.
- **Fix — CalibratorOn/CalibratorOff port-mutex contention** (AlpacaCore): both now dispatch to a background thread, mirroring the existing cover-task pattern, so the ASCOM initiator returns immediately with `CalibratorChanging`/`CalibratorState` correctly reporting `NotReady` while the change is in flight — regardless of whether the shared port is busy with a cover move. Also matches ConformU's own documented guidance for `CalibratorOn`.
- **Fix — clang-tidy `performance-unnecessary-value-param`**: `start_cover_task`'s `fn` parameter changed to `const std::function<void()>&`.
- **Documentation**: `AGENTS.md` (Gemini vendor notes, protocol quirks, the mutex-contention lesson), `SUPPORTED-DRIVERS.md` (new table row + driver notes), `docs/architecture.md` (vendor table), `CHANGELOG.md` (`[3.2.0] - UNRELEASED`).

## Test plan
- [x] Local CI pre-flight green on real Linux arm64 hardware: `clang-format`, unicode scan, `run_all_tests.sh` (vendors OFF: 32/32 + vendors ON: 339/339 AlpacaCore, 345/345 AlpacaHTTP), `clang-tidy`, `cppcheck`, `javascript` (shellcheck/zizmor skip — no matching files changed)
- [x] Gemini unit tests pass (146 assertions, 25 test cases across Lite + v2 + focuser)
- [x] ConformU 4.4.0 passes on Linux arm64 (0 errors, 0 issues, 0 timing issues), including after the CalibratorOn fix
- [x] Web UI configuration verified against real hardware (device registration, connect, cover open/close, calibrator on/off with varying brightness)

## ConformU results
- **arm64**: `AlpacaCore/conformu/Gemini/Astro Automatic FlatPanel v2/Linux-arm64.txt`


Knetus56:

joeytroy:
@Knetus56  thanks for adding to the project!

Knetus56:
@joeytroy pushed fixes for both findings from the review:
- `disconnect()` now routes through `start_connection_task(false)` (same pattern as `ZWOTelescopeDriver`) instead of calling `set_connected(false)` synchronously, so it can no longer block the HTTP handler for up to 30s behind an in-flight cover/calibrator wire command.
- Added a comment explaining why `start_calibrator_task()` intentionally joins an in-flight command instead of rejecting it like `start_cover_task()` does.

Note: the review workflow failed on this push with `Actor does not have write permissions to the repository" — looks like it only re-triggers automatically under your permissions, not mine as the fork contributor. Could you kick off another review when you get a chance?

joeytroy:
@Knetus56 working on it give me a few

Knetus56:
Pushed a fix for the `HaltCover` finding: it was latching `cover_halted_` unconditionally, permanently pinning `CoverState` to `Unknown` if called while the cover was already idle. Now only latches it when a move is actually in flight. The other point (calibrator task blocking behind an in-flight cover move) was flagged as a documented, non-blocking tradeoff, so no change there.

Knetus56:
Pushed a fix for the remaining calibrator-join finding: the reap of a still-in-flight previous calibrator command now happens on the new background thread (via a moved `std::thread` handle) instead of joining inline on the caller's thread, so `CalibratorOn`/`CalibratorOff` always return immediately regardless of whether a prior command is contended by a concurrent cover move. Successive commands still execute in submission order. Verified: rebuilds clean, Gemini unit tests pass (146 assertions), full suite passes (489 assertions).

Knetus56:
Pushed a fix for the `CalibratorChanging` finding: replaced the single `calibrator_changing_` bool with an `atomic<int>` pending-command count. Each background thread increments it on start and decrements on exit, so `NotReady`/`true` is reported until the actual tail of the chain finishes, regardless of which thread happens to complete first. Also reconciled the stale "NOT confirmed against real Rev2 hardware / provisional" header comments in `gemini_flatpanel_protocol_wrapper.h" now that ConformU has validated it. Verified: rebuilds clean, Gemini unit tests pass (146 assertions), full suite passes (489 assertions).

joeytroy:
@Knetus56 when ever you want to push I will leave it to you.

Knetus56:
@joeytroy CI is all green and it's mergeable — I still don't have push/merge rights on the repo, so could you merge this one when you get a chance?

joeytroy:
@Knetus56  There is still one issue found

One remaining concurrency issue found:

### `halt_cover()` has a TOCTOU race on `cover_halted_` (AlpacaCore/src/vendors/gemini/gemini_flatpanel_driver.cpp:645-650)

```cpp
void halt_cover() override {
    ensure_connected();
    if (cover_in_flight_.load()) {
        cover_halted_.store(true);
    }
}
```

Knetus56:
@joeytroy Fixed both — CoverState now self-heals via a live read instead of sticking at Unknown after HaltCover (even across disconnect/reconnect), and both start_cover_task()/start_calibrator_task() now roll back cleanly if std::thread's constructor throws (the calibrator path held the previous thread by shared_ptr instead of moving it, so a failed spawn can't call std::terminate() on a still-joinable thread). CI is green again and the automated review came back ✅ Approved with no outstanding issues. Ready to merge whenever you get a chance.

joeytroy:
@Knetus56 looks good, can we make one fix.

1. Stale text in index.html (~line 1538): the v2 help text says the command set is "not confirmed against real hardware" — but the PR body, AGENTS.md, and SUPPORTED-DRIVERS.md all say it was ConformU-validated on a real unit (firmware 408). This paragraph was clearly written pre-hardware and never updated. If you can update it before merge; it would confuse users.

One other question I have, I assume the when the user chooses the Flat panel, the do not need to put in any /dev/ttyUSB0 that Alpaca auto finds it? If not please fix that as all hardware should auto find.

Knetus56:
@joeytroy Fixed in 193ea7c:

1. Updated the stale v2 help text in `index.html` — it now says the cover/status command set was ConformU-validated against real hardware (firmware 408), matching the PR body/AGENTS.md/SUPPORTED-DRIVERS.md.

2. Auto-detect already works for v2 the same way it does for every other Gemini device — no fix needed there. The v2 config section has the same "Connection Type: Auto-detect (recommended)" dropdown and Panel Index field as the Lite model, and on the backend it goes through the identical `enumerate_gemini_flatpanel_ports()` port-scan (shared with the Lite model, including the same index namespace) rather than requiring a manual `/dev/ttyUSB0`. Users only need to type a port path if they explicitly switch Connection Type to "Serial (USB) — manual port".

PR #171, opened 2026-08-03
Validate ZWO AM5N ConformU 4.4.0 + park/tracking/connect timing fixes

zeroastroboy:
## Summary
- **ZWO AM5N mount validated** with ConformU 4.4.0 on Linux arm64 (USB): **0 errors, 0 issues, 0 timing issues**.
- Fixed four ZWO mount driver defects surfaced during validation: park-complete inference defeated by the poll-thread cache, `Tracking=true` while parked not throwing, `connected_` published before cache warm-up (cold FAST reads), and the `park()` method's latency versus the STANDARD 1.0 s target (kept at ~0.8 s).

## Changes
- **ZWO Mount Driver** (AlpacaCore): park-complete inference no longer defeated by the poll thread pinning `park_state_cached_=false` during active parks — the inferred-parked state sticks until explicit unpark, so ConformU's `Parked:` suite (slew/sync/tracking while parked) sees the mount as parked.
- **ZWO Mount Driver** (AlpacaCore): `set_tracking(true)` now throws `InvalidWhileParked` via `ensure_not_parked()` before its tracking-state fast path (matches SynScan/iOptron).
- **ZWO Mount Driver** (AlpacaCore): `connected_` is published only after `refresh_cached_values()` warms the telemetry caches; no-op reconnects keep the caches, so the first FAST reads after connect don't pay live serial round-trips (~95 ms each on the AM5N's USB-ACM link).
- **ZWO Mount Driver** (AlpacaCore): park state is warmed by the poll thread so `DeviceState`'s `AtPark` stays under the FAST 0.1 s target; the Park method stays under the STANDARD 1.0 s target (~0.8 s).
- **ConformU Validation**: `AlpacaCore/conformu/ZWO/AM5N/Linux-arm64.txt` (ConformU 4.4.0).
- **Documentation**: `SUPPORTED-DRIVERS.md` AM5N row → ✓, driver notes updated; `CHANGELOG.md` entry under UNRELEASED.

## Test plan
- [x] Unit tests pass (65 assertions in 11 ZWO telescope test cases)
- [x] ConformU 4.4.0 passes on Linux arm64 (USB) — 0 errors, 0 issues, 0 timing
- [x] Park / Unpark / Parked-while-slew / Tracking-while-parked behaviors verified

## ConformU results
- **AM5N (Linux arm64, USB)**: `AlpacaCore/conformu/ZWO/AM5N/Linux-arm64.txt` — 0 errors, 0 issues, 0 timing issues.


zeroastroboy:
Thanks for the review! The `:Sp01` note was correct — that change was added and then reverted during development, so the final diff has no `park()` change. I've updated the CHANGELOG and PR description to remove the stale claim. The park latency fix is reflected in the ConformU result (Park completes in ~0.8 s).

zeroastroboy:
Thanks for the detailed review! All points addressed:

1. **InProgress park-cache stomp** — fixed: the poll-thread warm no longer clears `parked_cached_` on `:Gps=InProgress` (the inference may have already set it true; stomping it would defeat park completion on AM3/AM5/AM7 firmware that keeps reporting InProgress after the mount stops). Now `InProgress` leaves both caches untouched, matching the intent.

2. **Auto-detect scope** — I've added unit tests covering `ConnectionType::Auto` (distinct transport value), `ZWODeviceInfo` defaults, `enumerate_zwo_mounts()` returning probed devices with non-empty model names, and clean connect-failure with no reachable mount (76 assertions / 13 ZWO test cases now). It stays in this PR per the feature request, but is now tested.

3. **`gethostbyname`** — `probe_network_mount` now uses thread-safe `getaddrinfo`.

4. **Singleton-mutex assumption** — documented in `enumerate_zwo_mounts()` that probes intentionally bypass the wrapper mutex because the singleton holds at most one physical ZWO connection at a time (same pattern as iOptron/SynScan).

zeroastroboy:
Thanks! Both points addressed:

1. **Unpark stale park_state_cached_** — fixed: `unpark()` now also resets `park_state_cached_`/`park_state_at_` (mirroring `find_home()`), so a fresh AtPark=true cache can no longer make `set_tracking(true)`/slew/sync throw `InvalidWhileParked` right after a successful unpark.

2. **probe_network_mount recv() hang** — this was fixed in the previous push: the probe now sets `SO_RCVTIMEO` on the socket before the receive loop, so a peer that accepts TCP but never answers :GVP fails within the probe timeout instead of hanging (the comment likely reviewed the commit before that fix landed).

Re-validated with these fixes: ConformU 4.4.0 clean (0 errors, 0 issues, 0 timing on Linux arm64/USB). Report refreshed.

zeroastroboy:
Fixed — good catch!

- **No-op reconnect wiping parked_cached_**: `reset_session_state_for_connect(keep_telemetry_caches=true)` now preserves `parked_cached_` (the sticky inferred-parked flag). It is only cleared on a full (fresh) connect, so a redundant Connected=true while parked can no longer silently un-park the driver once the short park_state_cached_ TTL elapses.
- **fcntl in probe_network_mount**: now checks the F_SETFL return and fails the probe on error.

The pre-existing `connect_network()` gethostbyname is noted as a follow-up (out of this PR's scope).

zeroastroboy:
Both edge cases fixed:

- **unpark_in_progress_ exception safety**: the :Spu round-trip is now inside try/catch and the guard is cleared on every exit path (success, rejected-unpark, and thrown serial/mount error) before rethrow, so a failed unpark can no longer permanently disable the poll thread's park-state cache warming.
- **caches_ready_ stuck false**: the fresh-connect path now wraps sync_mount_time_if_configured() in try/catch (mirroring the no-op reconnect branch), so a transient serial error during time/site sync cannot leave connected_=true with caches_ready_=false forever.

zeroastroboy:
Both points addressed:

1. **Sticky parked_cached_ tradeoff** — confirmed intentional and required: the AM5N's :Gps cannot be trusted to report parked (it returns NotParked/0 even while physically parked, which is exactly why the stationary-after-:hP inference exists), so allowing a live NotParked read to clear the cache would immediately re-break ConformU's Parked: suite. The state is cleared by the driver's own unpark() (and by a full disconnect/reconnect). Out-of-band unpark (hand controller) is the documented recovery path via the driver unpark().

2. **Auto-detect connect race** — fixed: connect() is refactored into a lock-free dispatcher + connect_locked() helper, and the ConnectionType::Auto branch now holds mutex_ for the whole probe+connect sequence. A concurrent explicit Serial/Network connect can no longer establish a link that the auto path silently tears down.

PR #176, opened 2026-08-05
Fix Gemini flat panel auto-detect missing ttyACM devices

Knetus56:
## Summary
- Fixes Gemini flat panel (Lite + v2) auto-detect so it also finds the panel on `/dev/ttyACM*`, not just `/dev/ttyUSB*`, when `/dev/serial/by-id` is unavailable.
- The panel's tested controller is an Espressif native USB-serial/JTAG stack, which Linux's `cdc_acm` driver exposes as `/dev/ttyACMn` — the by-id-absent fallback previously only probed `ttyUSB0..9` and would silently never find a genuine panel in that situation.

## Changes
- **Gemini Flat Panel Protocol Wrapper** (AlpacaCore `gemini_flatpanel_protocol_wrapper.cpp`): `enumerate_gemini_flatpanel_ports()`'s no-by-id fallback now probes both `/dev/ttyUSB0..9` and `/dev/ttyACM0..9`, matching the existing pattern already used in `zwo_mount_protocol_wrapper.cpp`.
- **Documentation**: CHANGELOG.md — new `[3.2.1] - UNRELEASED` / `### Fixed` entry (patch-level bump per versioning policy; `[3.2.0]` collapsed).

## Test plan
- [x] Local CI pre-flight green on arm64 (clang-format, Unicode scan, build+test vendors OFF/ON incl. Gemini unit tests, clang-tidy, cppcheck)
- [x] Full AlpacaCore unit test suite passes (341/341 vendors-ON, all Gemini Flat Panel/v2/Focuser cases green)
- [x] Deployed to a real arm64 test SBC and verified live: Gemini Astro Automatic FlatPanel v2 (firmware 408) connects, reports status, and closes its cover correctly through the rebuilt service
- [ ] ConformU — not re-run; this PR doesn't touch validated command paths, only port discovery

### DIFF
--- a/AlpacaCore/src/vendors/gemini/gemini_flatpanel_protocol_wrapper.cpp
+++ b/AlpacaCore/src/vendors/gemini/gemini_flatpanel_protocol_wrapper.cpp
@@ -237,13 +237,25 @@ std::vector<GeminiFlatPanelPortInfo> enumerate_gemini_flatpanel_ports() {
 #ifndef _WIN32
     const std::filesystem::path serial_by_id("/dev/serial/by-id");
     if (!std::filesystem::exists(serial_by_id)) {
-        for (int i = 0; i < 10; ++i) {
-            std::string port = "/dev/ttyUSB" + std::to_string(i);
-            if (std::filesystem::exists(port)) {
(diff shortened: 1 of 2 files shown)


Knetus56:

PR #194, opened 2026-08-08
Add clock sync for internet-less SBCs (sync-clock.sh + troubleshooting)

zeroastroboy:
## Summary
- Adds a **clock-sync solution for internet-less SBCs**: `scripts/sync-clock.sh` pushes a workstation's current time to an SBC over SSH (with optional `--rtc` persist), plus a troubleshooting guide entry.
- Solves the "device clock resets to a stale time after reboot" problem that makes Alpaca timestamps (and ConformU `LastExposureStartTime` checks) wrong on headless SBCs with no NTP reachable and a dead/unset hardware RTC.

## Changes
- **Script**: `scripts/sync-clock.sh` — pushes time over SSH (`ssh <target> "sudo date -s ..."`), `--rtc` also runs `hwclock -w`; works over any SSH path (direct cable, WiFi AP, LAN).
- **Docs**: `docs/troubleshooting.md` — new "Device clock resets to a stale time after reboot" section covering the full fix (set clock manually, install systemd-timesyncd/chrony, `timedatectl set-ntp true`, `hwclock -w`, verify with `timedatectl status`), and the SSH push as the no-internet fallback.
- **CHANGELOG**: entry under UNRELEASED.

## Why
A headless SBC with no internet has no NTP source; if its RTC is unset or battery-dead, the kernel boots with a stale value (e.g. 2017/2025). This surfaces as wrong Alpaca timestamps and ConformU `LastExposureStartTime` accuracy failures. The SSH push is the simplest no-internet time source: any machine that can SSH to the SBC can also set its clock.

## Test plan
- [x] Verified live: pushing time from the workstation to a Debian 13 SBC matches the clocks exactly
- [x] `--rtc` persists to the hardware RTC (`hwclock -w`)
- [x] NTP path (systemd-timesyncd + `timedatectl set-ntp true`) shows `System clock synchronized: yes`

## Web UI Sync Time button (added)

In addition to the CLI script, the web portal now has a **Sync Time** button (top-right server actions) that sets the SBC's clock from the browser's time:

- **Endpoint**: `POST /management/v1/synctime` — accepts `{"Epoch": <unix-seconds>}`, validates the 2000–2100 range, calls `clock_settime(CLOCK_REALTIME, ...)`
- **Permission**: `alpacabridge` systemd unit grants `CAP_SYS_TIME` (ambient, compatible with `NoNewPrivileges=true`) so the non-root service can set the clock
- **UI**: button styled to match the theme, with confirm dialog + success/error feedback
- **Verified live**: clock synced exactly to the browser time; invalid/missing epochs rejected; GET → 405


zeroastroboy:
All five points addressed:

1. **Stale epoch after confirm** — fixed: `syncTime()` now captures `Date.now()` *after* the `confirm()` dialog returns, so the clock is set to when the user confirmed (confirm() blocks, so the old code sent a stale click-time epoch).
2. **sync-clock.sh NTP comment** — removed the misleading "installs an NTP client" claim; the script only pushes time (NTP setup remains a documented manual step).
3. **--rtc as sole arg** — `--rtc` alone now falls back to the default host instead of being swallowed as the target.
4. **Security / threat model** — documented in the handler: the endpoint is intentionally unauthenticated like restart/shutdown (LAN-only web UI, trusted-network model), the epoch is sanity-bounded 2000–2100 UTC, and deployments on untrusted networks should firewall the management port. Confirmed this matches the project's existing management-endpoint threat model.
5. **SSH host-key checking** — `StrictHostKeyChecking=accept-new` (auto-accept unknown, reject changed keys) instead of unconditional `no`.

joeytroy:
Response to the latest review round:

1. **Unauthenticated endpoint / no rate limiting** — accepted trade-off, consistent with the existing restart/shutdown management endpoints and documented in the handler comment (LAN-trust model; firewall the management port on untrusted networks). No change.
2. **`CapabilityBoundingSet` replaces the set** — confirmed intentional: the service runs as a non-root user and previously needed (and had) no ambient capabilities, so restricting the bounding set to `CAP_SYS_TIME` is pure hardening. Noted for the future: any driver that later needs another capability must extend this line. Verified on hardware (OrangePi, Debian 13 arm64): `CapEff/CapAmb/CapBnd = cap_sys_time` only, and the clock-set works under `NoNewPrivileges=true`.
3. **Delta bound on clock jumps** — deliberately not added: the core use case is a dead/unset RTC resetting the clock to a years-old date on boot, so a legitimate first sync routinely jumps years. Any delta bound that still covers that case would be too wide to add meaningful protection beyond the existing 2000–2100 range check.
4. **Transient GET failure blanking the live clock** — fixed in 86cbc89: the clock keeps ticking off the last known offset until a re-sync succeeds.

PR #201, opened 2026-08-12
Fix QHY camera exposure hang and HDR readout timeout

htly9981:
## Summary
- Fixes a bug where QHY miniCam8M exposures could hang indefinitely with no
  error and sometimes take the camera offline
- Root cause: `ExpQHYCCDSingleFrame` and `GetQHYCCDSingleFrame` were called
  from different threads; QHY's own SDK sample keeps both calls on the same
  thread and this driver didn't
- Also fixes a false-failure watchdog timeout on the "Linearity HDR" readout
  mode, which legitimately takes ~64s to download regardless of exposure
  duration or `USBTRAFFIC`/`CONTROL_HDR` settings

## Changes
- **QHY Camera Driver** (AlpacaCore): `start_exposure()` now runs the entire
  Exp→Get sequence on one worker thread, matches the SDK sample's call order
  (`SetQHYCCDResolution` before `SetQHYCCDBinMode`, `GetQHYCCDMemLength`
  fetched after arming), and adds the sample's 1s settle delay before the
  download read. The exposure worker join is now bounded (2s timeout +
  detach fallback) so a wedged download can no longer hang
  `StartExposure`/`StopExposure`/`Disconnect` forever. Temp-control and
  telemetry background threads now skip their SDK polling while an exposure
  is in progress.
- **QHY SDK Wrapper** (AlpacaCore): every SDK call (temp control, telemetry,
  CFW, gain/offset, exposure arm/download, etc.) is now serialized against
  the same physical handle via a per-handle mutex, so concurrent callers
  can no longer interleave calls into the vendor SDK on one handle.
  `cancel_exposure()` deliberately stays unserialized since it needs to be
  able to interrupt a blocked `GetQHYCCDSingleFrame` call from another
  thread.
- **Watchdog deadline**: the exposure deadline is now re-derived from the
  actual buffer size once `GetQHYCCDMemLength()` returns (extending, never
  shrinking, the existing flat 60s margin), so it scales with the readout
  mode instead of assuming every mode transfers at the same rate — fixes
  "Linearity HDR" mode being killed as falsely "stuck" mid-download.
- **Documentation** (SUPPORTED-DRIVERS.md, CHANGELOG.md): documents the
  Linearity HDR ~64s download behavior as a known SDK-internal limitation
  (reported to QHY) in the QHY Driver Notes section, so future contributors
  don't mistake it for a driver bug; adds a `[3.4.1] - UNRELEASED`
  CHANGELOG entry for both fixes.

## Test plan
- [x] Local CI pre-flight green: clang-format, unicode scan, build+test
      (vendors OFF and ON, 2420+ assertions), clang-tidy, cppcheck
- [x] Verified on real miniCam8M hardware: repeated Full Resolution
      exposures (4s, 60s) complete and download successfully, no hangs
- [x] Verified on real hardware: Linearity HDR mode exposures (single and a
      3-frame sequence) complete around the expected ~65s mark without a
      watchdog false-failure


htly9981:

joeytroy:
@htly9981 Thanks for this — two CI checks are red, but only one needs a change from you:

**`clang-format (changed lines)` — needs a one-line reformat.** `clang-format` wants the log call in `AlpacaCore/src/vendors/qhy/qhy_camera_driver.cpp` (~line 1395) wrapped after the first argument instead of mid-expression:

```cpp
// current (fails)
ALPACA_LOG_WARN("QHY", "cancel_exposure after get_mem_length failure failed: " +
                 std::string(e.what()));

// what clang-format wants
ALPACA_LOG_WARN("QHY",
                "cancel_exposure after get_mem_length failure failed: " + std::string(e.what()));
```

Easiest fix on your branch:

```bash
git-clang-format 1e58daeb8dc75e9b4b5bac50fa1e9fc38938d91e
git commit -am "clang-format: wrap ALPACA_LOG_WARN in qhy_camera_driver"
git push
```

(`1e58daeb` is the merge base with `main` — this only touches your changed lines.)

**`review` — nothing for you to do.** The automated review bot failed to post its comment (transient); I've re-run that job and it should go green on its own.

htly9981:
@joeytroy All review feedback addressed — please kick off another round when you have a chance.

Latest two findings from the previous round:
- **Stale-worker cancel ordering**: the `mem_length == 0` exposure-worker branch now checks `exposure_superseded` *before* calling `cancel_exposure()`, so a detached zombie worker can no longer abort a newer, legitimate exposure on the same handle.
- **Whole-driver deadlock guard**: `ensure_not_exposing_locked()` now also rejects while the exposure worker thread might still be alive (not just while `exposure_status_ == Working`), closing the window where a wedged download holding `call_mutex` could cause every other setter — and the recovery path itself — to block on `mutex_` forever.

Both fixes committed in `80b3112`, plus a `0290d9d` re-validation: real miniCam8M hardware, ConformU 4.5.0, 0 errors / 0 issues / 0 timing issues, with the AlpacaBridge TRACE log confirming no false "camera busy" errors on the normal exposure→setter sequence.

htly9981:
@joeytroy Another round done — please re-request review when you get a chance.

Latest finding: `connect()` could open a second, independent handle to the same physical camera if a prior exposure worker got detached (2s join timeout) rather than joined, since `disconnect()` unconditionally erases the SDK wrapper's handle-map entry regardless of whether the join actually succeeded. Fixed by having `connect()` check `exposure_thread_running_` first and refuse to reopen ("still finishing" error) instead of risking a dual-handle open. Also fixed the stale `disconnect()` comment that claimed the worker was "already cancelled + joined" (not true in the detach case).

Committed in `50540ef`. Local CI (build+test both vendor configs, clang-tidy, cppcheck, and clang-format verified directly via `git-clang-format`) all clean.

htly9981:
@joeytroy Another round — please re-request review when convenient.

**Fixed**: TOCTOU race on `exposure_superseded` across all four check sites in the exposure worker lambda (`start_single_frame` failure, `mem_length == 0`, deadline-extension, final `get_single_frame` completion). Each only checked the flag *before* acquiring `mutex_`; since `load()` and the lock aren't one atomic step, a newer generation could finish arming/running/publishing its own state in that gap, and the stale worker would clobber it on acquiring the lock. Each site now re-checks `exposure_superseded` immediately after acquiring `mutex_` too — while deliberately keeping the original pre-lock check, since that one guards a different hazard (`this` itself may already be destroyed by the time this stale worker resumes; the pre-lock read never touches `this`, which is what makes it safe to even attempt the lock).

**Not changed**: the CFW-blocks-during-download note — that's the same accepted trade-off already documented in this PR (un-serializing would reopen the unbounded-hang risk the shared `call_mutex` exists to prevent).

Committed in `a1c17f0`. Local CI (build+test both vendor configs, clang-tidy, cppcheck, clang-format verified directly via `git-clang-format`) all clean.

htly9981:
@joeytroy Another round — please re-request review when convenient.

**Fixed**: the dual-handle-on-reconnect guard from the previous round only protected the camera driver's own `connect()` (it checked `exposure_thread_running_`, a private member of that driver instance). The paired CFW driver shares the same physical handle and calls `QHYSDKWrapper::open_camera()` independently with no visibility into that member — so reconnecting the CFW alone while a zombie camera exposure worker was still blocked in `GetQHYCCDSingleFrame` could still open a second physical handle.

Moved the check into `open_camera()` itself, keyed by `camera_id` via a new `register_exposure_worker()` call (`start_exposure()` registers the same `exposure_thread_running_` shared_ptr, stored separately from the handle map so it survives `close_camera()` erasing that entry). Both `connect()` paths (camera and CFW) share this one `open_camera()`, so the CFW driver needed zero code changes.

Committed in `e18b286`. Local CI (build+test both vendor configs, clang-tidy, cppcheck) clean; clang-format actually caught two real formatting nits this round (include order, one over-wrapped signature) — fixed and re-verified via `git-clang-format`.

htly9981:
@joeytroy Another round — please re-request review when convenient.

**Fixed**: `start_exposure()`'s reap-the-previous-worker block attempted `cancel_exposure()` + a bounded `join_exposure_thread()`, but never re-checked `exposure_thread_running_` afterward before falling through to the `call_mutex`-guarded setters (`set_param(EXPOSURE)`, `set_resolution`, `set_bin_mode`, `set_bits_mode`). If `cancel_exposure()` failed to unblock a genuinely wedged `GetQHYCCDSingleFrame`, those setters would block on `call_mutex` indefinitely — reintroducing the original hang one level down. Also worse than it looked: when a *prior* reap had already detached the zombie, `join_exposure_thread()` (and its 2s grace period) was skipped entirely, so there was no wait at all before hitting the mutexed setters.

`start_exposure()` now re-checks `exposure_thread_running_` right after the reap attempt and throws the same "camera busy, try again" error every other setter already uses via `ensure_not_exposing_locked()`.

Committed in `9f372db`. Local CI (build+test both vendor configs, clang-tidy, cppcheck) clean; clang-format caught one real wrap-style nit again — fixed and re-verified via `git-clang-format`.

htly9981:
@joeytroy Another round — please re-request review when convenient.

**Fixed**: the watchdog-deadline-extension block (`if (readout_mode_ != 0) { ... }`, run right after a successful `get_mem_length()`) read `this->readout_mode_` directly with no `exposure_superseded` check beforehand — unlike every other post-blocking-call touch of `this` in the same worker lambda. `get_mem_length()` is itself a `call_mutex`-guarded SDK call that can take longer than expected, long enough for a concurrent `join_exposure_thread()` elsewhere to hit its 2s timeout and detach this worker — and, via the destructor, finish tearing `this` down — while it was still blocked there or in the preceding 1s settle sleep. Added the same `exposure_superseded->load()` guard used everywhere else in this lambda, right before the `readout_mode_` read.

**Not changed** (per the review's own "no action needed" notes): the permanent-unusable-until-restart ceiling if `GetQHYCCDSingleFrame` truly never returns is accepted by design (documented in-code); `mem_length == 0` surfacing asynchronously instead of synchronously is an accepted minor API shift, essentially a hardware-fault case.

Committed in `d29e3ab`. Local CI (build+test both vendor configs, clang-tidy, cppcheck, clang-format via `git-clang-format`) all clean.

PR #204, opened 2026-08-16
feat(touptek): cooling-fan control API + persist fan speed across connects

zeroastroboy:
## Problem

ToupTek cameras with a cooling fan (e.g. the ATR2600M) run the fan whenever the SDK handle is open. The SDK supports `TOUPCAM_OPTION_FAN` (0 = off, [1, max] = speed) and the wrapper already exposed `get_fan`/`put_fan`, but nothing surfaced it — the fan could only be silenced from a raw SDK handle, and the setting resets to the hardware default on every camera open.

## Fix

- `CameraDriver` gains vendor fan virtuals (`get_supports_fan` / `get_fan_speed` / `set_fan_speed` / `get_max_fan_speed`), defaulting to NotImplemented (Alpaca 1025) so the other drivers are untouched.
- `ToupTekCameraDriver` implements them via the wrapper and **remembers the last set speed, re-applying it on every connect** (the setting is volatile — it resets on each SDK open).
- Router exposes the vendor extension endpoints:
  - `GET /api/v1/camera/{n}/fan` → current speed (0 = off)
  - `PUT /api/v1/camera/{n}/fan?FanSpeed=0` → set speed
  - `GET /api/v1/camera/{n}/fanmaxspeed` → max speed
- An arm64 `.deb` build workflow so the patched bridge can be installed without a local arm64 toolchain.

## Verification

- Built via the new workflow on GitHub's native arm64 runner → `.deb` installed on the rig
- Live: `GET /fan` → 1, `PUT FanSpeed=0` → success, `GET /fan` → 0, and the fan **stays off across a camera reconnect** (the re-apply on connect works)

### DIFF
--- a/.github/workflows/build-deb.yml
+++ b/.github/workflows/build-deb.yml
@@ -1,0 +1,48 @@
+name: Build .deb (arm64)
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - 'feat/**'
+
+jobs:
+  build-deb:
+    name: dpkg-buildpackage (arm64)
+    runs-on: ubuntu-24.04-arm
+    steps:
+      - name: Checkout
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v4.x pinned (immutable ref)
+        with:
+          persist-credentials: false
+
+      - name: Install build dependencies
+        run: |
+          sudo apt-get update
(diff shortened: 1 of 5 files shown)


zeroastroboy:
Superseded — the fan is already exposed via the ToupTek Thermal Switch device (TOUPCAM_OPTION_FAN as a switch element); no new bridge route needed.

